PR #7, opened 2026-01-13
add explicit timeout when sending watchdog requests

Ivansete-status:


fcecin (approved):

NagyZoltanPeter (approved):
Great, assuming it's an exit sign for the user of the lib, in 20sec the node would be already kicked out from the mesh.

PR #14, opened 2026-05-01
use fixed array of ctx to avoid consuming all fds

### LINKED ISSUE #4: leaked file descriptors
While testing nim-sds v0.2.3 in status-go, it's been detected that `epoll` file descriptors get leaked.

See discussion in: https://github.com/status-im/status-go/issues/7216

In order to overcome that, the following is suggested, but we still need to understand why the fd leakage happens:

- https://github.com/logos-messaging/nim-sds/pull/40
- ~~https://github.com/logos-messaging/sds-go-bindings/pull/9~~
- https://github.com/status-im/status-go/pull/7227

@igor-sirotin @arnetheduck kindly extend the issue description with anything I might have missed

Ivansete-status:
Use fixed array of 32 context to allow reuse them and hence, prevent app from running out of file descriptors.
Notice that this is mostly needed for testing purposes. For example, we detected an issue when dealing with Go unit tests that started/stopped several contexts, without recycling, and then the test process run out of FDs.

closes https://github.com/logos-messaging/nim-ffi/issues/4

p.s.: there are some other minor changes like enhancing the error handling, use of wait signal in watchdog to allow fast shutdown (and make tests faster.)


ffi/ffi_context.nim
```diff
@@ -164,15 +172,17 @@ proc processRequest[T](
     try:
       await retFut
     except CatchableError as exc:
```
gmelodie: Not the code that this PR introduces, but we should avoid `CatchableError` as much as possible, use specific errors instead, and give more context to error messages

ffi/ffi_context.nim
```diff
@@ -110,13 +112,19 @@ proc watchdogThreadBody(ctx: ptr FFIContext) {.thread.} =
...
-    # Give time for the node to be created and up before sending watchdog requests
-    await sleepAsync(WatchdogStartDelay)
-    while true:
-      await sleepAsync(WatchdogTimeinterval)
-
-      if ctx.running.load == false:
-        debug "Watchdog thread exiting because FFIContext is not running"
+    # Give time for the node to be created and up before sending watchdog requests.
+    # waitSync returns early if watchdogStopSignal fires (i.e. on destroy).
+    let startWait = ctx.watchdogStopSignal.waitSync(WatchdogStartDelay)
```
gmelodie: nit: `startWait` and `intervalWait` give me the impression that they're `Duration`s, maybe we could change the names? I think both could be named `stopSignal`
```suggestion
    let stopSignal = ctx.watchdogStopSignal.waitSync(WatchdogStartDelay)
```

ffi/ffi_context.nim
```diff
@@ -208,42 +218,56 @@ proc ffiThreadBody[T](ctx: ptr FFIContext[T]) {.thread.} =
 
   waitFor ffiRun(ctx)
 
-proc cleanUpResources[T](ctx: ptr FFIContext[T]): Result[void, string] =
-  defer:
-    freeShared(ctx)
+proc closeResources*[T](ctx: ptr FFIContext[T]): Result[void, string] =
```
gmelodie: nit
```suggestion
proc cleanup*[T](ctx: ptr FFIContext[T]): Result[void, string] =
```

ffi/ffi_context.nim
```diff
@@ -208,42 +218,56 @@ proc ffiThreadBody[T](ctx: ptr FFIContext[T]) {.thread.} =
...
 
-proc createFFIContext*[T](): Result[ptr FFIContext[T], string] =
-  ## This proc is called from the main thread and it creates
-  ## the FFI working thread.
-  var ctx = createShared(FFIContext[T], 1)
+proc cleanUpResources[T](ctx: ptr FFIContext[T]): Result[void, string] =
+  ## Full cleanup for heap-allocated contexts: closes all resources and frees memory.
+  defer:
+    freeShared(ctx)
+  return ctx.closeResources()
```
gmelodie: We could have a single `cleanup` proc and add a `freeCtx: bool = false` (or default to `true`, if that makes more sense)

ffi/ffi_context.nim
```diff
@@ -252,31 +276,51 @@ proc createFFIContext*[T](): Result[ptr FFIContext[T], string] =
...
+  let ffiSignaled = ctx.reqSignal.fireSync().valueOr:
     ctx.onNotResponding()
-    return err("error in destroyFFIContext: " & $error)
-  if not signaledOnTime:
+    return err("error signaling reqSignal in destroyFFIContext: " & $error)
+  if not ffiSignaled:
     ctx.onNotResponding()
     return err("failed to signal reqSignal on time in destroyFFIContext")
-
+  let wdSignaled = ctx.watchdogStopSignal.fireSync().valueOr:
```
gmelodie: `wdSignaled` is the same as above, should have the same name. I'd just called it `stop` or maybe `stopSignaled`, but call it that everywhere

ffi/ffi_context.nim
```diff
@@ -252,31 +276,51 @@ proc createFFIContext*[T](): Result[ptr FFIContext[T], string] =
...
+    return err("error signaling watchdogStopSignal in destroyFFIContext: " & $error)
+  if not wdSignaled:
+    return err("failed to signal watchdogStopSignal on time in destroyFFIContext")
```
```suggestion
    return err("error signaling stopSignal in destroyFFIContext: " & $error)
  if not wdSignaled:
    return err("failed to signal stopSignal on time in destroyFFIContext")
```

gmelodie (approved):

gmelodie (approved):

ffi/ffi_context.nim
Ivansete-status: Thanks! This will be covered by https://github.com/logos-messaging/nim-ffi/issues/18

examples/nim_timer/nim_timer.nim
```diff
@@ -72,7 +72,8 @@ proc nimtimerComplex*(
 # In a multi-file library, import all sub-modules first and call genBindings()
 # once, at the bottom of the top-level compilation-root file.
 # This call is a no-op unless -d:ffiGenBindings is passed to the compiler.
-genBindings() # reads -d:ffiOutputDir, -d:ffiNimSrcRelPath, -d:targetLang from compile flags
+genBindings()
+  # reads -d:ffiOutputDir, -d:ffiNimSrcRelPath, -d:targetLang from compile flags
```
gmelodie: nit: maybe this comment should be inside the definition of `genBindings`? Here it may cause confusion as to what it refers to

ffi/internal/ffi_macro.nim
```diff
@@ -1476,8 +1479,13 @@ macro ffiCtor*(prc: untyped): untyped =
       )
     )
 
+  let poolDecl = quote do:
```
gmelodie, line 1490: TIL about `quote`!

ffi/ffi_context.nim
```diff
@@ -358,54 +359,116 @@ proc createFFIContext*[T](): Result[ptr FFIContext[T], string] =
...
+type FFIContextPool*[T] = object
+  ## Fixed-size pool of FFI contexts. Avoids dynamic heap allocation per context
+  ## and bounds the total number of file descriptors consumed by ThreadSignalPtrs
+  ## to at most MaxFFIContexts * 2.
+  slots: array[MaxFFIContexts, FFIContext[T]]
+  inUse: array[MaxFFIContexts, Atomic[bool]]
+
+proc acquireSlot[T](pool: var FFIContextPool[T]): Result[ptr FFIContext[T], string] =
+  for i in 0 ..< MaxFFIContexts:
+    var expected = false
```
```suggestion
  const expected = false
  for i in 0 ..< MaxFFIContexts:
```

ffi/ffi_context.nim
```diff
@@ -358,54 +359,116 @@ proc createFFIContext*[T](): Result[ptr FFIContext[T], string] =
...
 
   if not exitedOnTime:
-    ## Event loop is blocked by a synchronous handler. Leak the thread and
-    ## ctx to avoid hanging the caller forever.
     ctx.onNotResponding()
     return err("FFI thread did not exit in time; leaking ctx to avoid hang")
 
   joinThread(ctx.ffiThread)
   joinThread(ctx.watchdogThread)
   ctx.cleanUpResources().isOkOr:
```
gmelodie: This line seems to be the only difference between the two `destroyFFIContext` implementations, maybe we could wrap the rest in another function and make the two impls call the new function? e.g.
```nim
proc newFn(ctx...) =
  unregisterCtx(cast[pointer](ctx))
  ...
  joinThread(ctx.ffiThread)
  joinThread(ctx.watchdogThread)

# then ...
  
proc destroyFFIContext*[T](
    pool: var FFIContextPool[T], ctx: ptr FFIContext[T]
): Result[void, string] =
  newFn(ctx)
  pool.releaseSlot(ctx)
  return ok()

# and ...

proc destroyFFIContext[T](ctx: ptr FFIContext[T]): Result[void, string] =
  newFn(ctx)
  ctx.cleanUpResources().isOkOr:
    return err("cleanUpResources failed: " & $error)
  return ok()
```

gmelodie (commented):
Besides the other comments, do we even use this at all? 
```nim
proc destroyFFIContext[T](ctx: ptr FFIContext[T]): Result[void, string] =
```
I see that tests introduced all use `pool.destroyFFIContext`

NagyZoltanPeter (approved):
Yepp, this is needed but I see a bit of discrepancy in between we reserve slots and store the ctx pointer in there while we have another store for ctx ptrs.
I suggest a bit of rework here, eliminate that `contextRegistry` for which we always do a lock at every api call and just use the slot registry for the same puropse - for validating the ctx ptr - reading slot registry does not need a lock.

Ivansete-status:
> Besides the other comments, do we even use this at all? 
```nim
proc destroyFFIContext[T](ctx: ptr FFIContext[T]): Result[void, string] =
```

@gmelodie - this proc is needed to clear the context. I renamed it in the following, to avoid confusion: 
b53f3cee1364d711fbefd38cb52ee324b91865bd

examples/nim_timer/nim_timer.nim
Ivansete-status: Addressed in 397150e6bc3b7d51c606950121b320369ec3d814

ffi/ffi_context.nim
Ivansete-status: Thanks but `compareExchange` in atomics requires a `var`:
```nim
proc compareExchange[T: not Trivial](location: var Atomic[T]; expected: var T;
                                     desired: T; order: MemoryOrder = moSequentiallyConsistent): bool
```

ffi/ffi_context.nim
Ivansete-status: I think it was confusing because both had the same name but one call the other.
The https://github.com/logos-messaging/nim-ffi/commit/b53f3cee1364d711fbefd38cb52ee324b91865bd may have enhanced that.

Ivansete-status:
> Yepp, this is needed but I see a bit of discrepancy ...

Wow! Great catch, thanks @NagyZoltanPeter !
I didn't properly solve the conflicts after rebasing and it got messy. I think it is better now. Addressed in dbc58520f4acba19898001b83279627cc1888848

Issue #17, opened 2026-05-08
add coverage for orc and refc


Ivansete-status:
This request comes from @NagyZoltanPeter, thanks!
We need to add new tests, and extend the current ffi.nimble, so that we make sure there are no issues with orc or refc.

gmelodie:
Finished in #20

PR #20, opened 2026-05-08
chore: run tests with refc

gmelodie:
Part of #17. Runs tests with refc and orc

### DIFF
--- a/ffi.nimble
+++ b/ffi.nimble
@@ -12,17 +12,23 @@ requires "chronos"
 requires "chronicles"
 requires "taskpools"
 
-const nimFlags = "--mm:orc -d:chronicles_log_level=WARN"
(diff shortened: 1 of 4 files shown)


ffi.nimble
```diff
@@ -12,17 +12,23 @@ requires "chronos"
 requires "chronicles"
 requires "taskpools"
 
-const nimFlags = "--mm:orc -d:chronicles_log_level=WARN"
+const nimFlagsOrc = "--mm:orc -d:chronicles_log_level=WARN"
```
Ivansete-status, line 15: That's great indeed!
I wonder if we should split the CI jobs in refc and orc

Ivansete-status (approved):
LGTM! Thanks for it! 🥳

PR #23, opened 2026-05-13
Start using CBOR

Ivansete-status:
### What

One CBOR blob per request, one per response. The C-exported wrapper now takes `(reqCbor: ptr byte, reqCborLen: csize_t)` instead of N `cstring` params. Success payloads are CBOR; error payloads stay as raw UTF-8.

### Why

- Strong typing on the wire (compile-time field walk, not a dynamic `JsonNode` graph).
- Transport-agnostic: same byte stream can later flow over a socket.
- Simpler memory model: one uniform `deleteRequest`, no per-type free.

### Changes

- **`ffi/cbor_serial.nim`** (new) — in-tree RFC 8949 codec (no new dep). Covers primitives, `seq`, `array`, `Option`, enum, `ptr T`, object/tuple. Replaces `ffi/serial.nim`.
- **`ffi/ffi_thread_request.nim`** — `FFIThreadRequest` owns one CBOR buffer; uniform `deleteRequest`; callback never receives a nil msg ptr (CBOR `null` on empty-OK, `"unknown error"` on empty-ERR).
- **`ffi/internal/ffi_macro.nim`** — `{.ffi.}` / `{.ffiCtor.}` / `{.ffiDtor.}` emit a `(reqCbor, reqCborLen)` C wrapper; per-proc machinery (`buildFfiDeleteReqProc`, `deleteReqContent`) removed. The user's Nim name now resolves to a natural `proc(lib, ...): Future[Result[T, string]]` — the C wrapper is an overload sharing the name via `{.exportc.}`. Sync `{.ffi.}` keeps its inline fast path but the Nim wrapper still returns `Future[...]` to preserve the source-level shape.
- **Codegen** — `ffi/codegen/rust.nim` uses `ciborium`; `ffi/codegen/cpp.nim` uses `nlohmann::json::to_cbor`/`from_cbor`. Both emit per-proc `<Proc>Req` structs.
- **Tests** — new `test_cbor_serial.nim` (round-trip per type + error cases) and `test_nim_native_api.nim` (calls `{.ffi.}` procs by their natural signature, including an `{.ffi.}` proc that `await`s another). Existing tests updated for the new ABI and CBOR responses.
- **Example** — `nim_timer` switched to `{.ffiDtor.}`; pre-existing `declareLibrary` mismatch fixed.

### Notes

- `ptr T` / `pointer` encoded as CBOR unsigned int (`cast[uint](p)`) — in-process only. A handle-table layer is needed before sending these across a network; not in scope in this PR.

### Verification

All tests pass under both `--mm:orc` and `--mm:refc`. `nimble genbindings_rust` and `nimble genbindings_cpp` produce well-formed bindings against the nim_timer example.


examples/nim_timer/cpp_bindings/nimtimer.hpp
```diff
@@ -114,24 +125,33 @@ struct FfiCallState_ {
...
+                s.bytes.assign(
+                    reinterpret_cast<const std::uint8_t*>(msg),
+                    reinterpret_cast<const std::uint8_t*>(msg) + len);
+            } else {
+                s.err.assign(msg, len);
+            }
+        }
         s.done = true;
         s.cv.notify_one();
     }
```
gmelodie: This function is a bit cryptic. To make it less so, we could move this block to a proper function and call it here. It'd be the same thing, but the function name alone would help understand the logic.

examples/timer/cpp_bindings/timer.hpp
```diff
@@ -144,8 +164,8 @@ inline std::string ffi_call_(std::function<int(FfiCallback, void*)> f,
     if (!fired)
         throw std::runtime_error("FFI call timed out after " + std::to_string(timeout.count()) + "ms");
     if (!state->ok)
-        throw std::runtime_error(state->msg);
-    return state->msg;
+        throw std::runtime_error(state->err);
```
gmelodie, line 652: Nice!

examples/nim_timer/cpp_bindings/nimtimer.hpp
```diff
@@ -157,15 +177,17 @@ inline std::string ffi_call_(std::function<int(FfiCallback, void*)> f,
 class NimTimerCtx {
 public:
     static NimTimerCtx create(const TimerConfig& config, std::chrono::milliseconds timeout = std::chrono::seconds{30}) {
-        const auto config_json = serializeFfiArg(config);
+        const auto req = NimtimerCreateCtorReq{config};
```
gmelodie: could it be this instead?
```suggestion
        const auto req = TimerReq{config};
```
ps: why is it called NimTimer instead of just timer?

examples/nim_timer/cpp_bindings/nimtimer.hpp
```diff
@@ -175,7 +197,7 @@ class NimTimerCtx {
 
     ~NimTimerCtx() {
         if (ptr_) {
-            nimtimer_destroy(ptr_);
+            nimtimer_destroy(ptr_, nullptr, nullptr);
```
gmelodie: can these be defaults? 
```suggestion
            nimtimer_destroy(ptr_);
```

not sure if C++ has default args, if not maybe polymorph the function?
```
nimtimer_destroy(ptr, arg1, arg2) {
...
}

nimtimer_destroy(ptr) {
  nimtimer_destroy(ptr, nullptr, nullptr);
}
```
same for other occurences

examples/timer/rust_bindings/src/api.rs
```diff
@@ -1,38 +1,55 @@
...
+        };
         state.payload = Some(if ret == 0 {
-            Ok(CStr::from_ptr(msg).to_string_lossy().into_owned())
+            Ok(bytes)
         } else {
-            Err(CStr::from_ptr(msg).to_string_lossy().into_owned())
+            Err(String::from_utf8_lossy(&bytes).into_owned())
         });
         cvar.notify_one();
     }
```
gmelodie: Analogous to my comment about the block in cpp, this could be a function instead

examples/timer/rust_bindings/src/api.rs
```diff
@@ -56,32 +73,37 @@ where
...
+        slice::from_raw_parts(msg as *const u8, len).to_vec()
+    };
     let value = if ret == 0 {
-        Ok(CStr::from_ptr(msg).to_string_lossy().into_owned())
+        Ok(bytes)
     } else {
-        Err(CStr::from_ptr(msg).to_string_lossy().into_owned())
+        Err(String::from_utf8_lossy(&bytes).into_owned())
     };
     let _ = tx.send(value);
```
gmelodie: This returns an error when fails, do we want to pop this up?

examples/timer/rust_bindings/src/api.rs
```diff
@@ -56,32 +73,37 @@ where
...
 
-async fn ffi_call_async<F>(f: F) -> Result<String, String>
+async fn ffi_call_async<F>(f: F) -> Result<Vec<u8>, String>
 where
     F: FnOnce(ffi::FfiCallback, *mut c_void) -> c_int,
 {
     let rx = {
-        let (tx, rx) = tokio::sync::oneshot::channel::<Result<String, String>>();
+        let (tx, rx) = tokio::sync::oneshot::channel::<Result<Vec<u8>, String>>();
         let raw = Box::into_raw(Box::new(tx)) as *mut c_void;
```
gmelodie: Some serious cool dark magic going on here. Could we extract this into a function?
```rust
fn into_raw_box_pointer(tx) -> *mut c_void {
    Box::into_raw(Box::new(tx)) as *mut c_void;
}

// and if we also need for rx

fn into_raw_box_pointer(rx) -> *mut c_void {
    Box::into_raw(Box::new(rx)) as *mut c_void;
}
```

ffi.nimble
```diff
@@ -34,9 +35,9 @@ task test_ffi, "Run FFI context integration tests under --mm:orc and --mm:refc":
   exec "nim c -r " & nimFlagsOrc & " tests/test_ffi_context.nim"
   exec "nim c -r " & nimFlagsRefc & " tests/test_ffi_context.nim"
 
-task test_serial, "Run serial unit tests":
-  exec "nim c -r " & nimFlagsOrc & " tests/test_serial.nim"
-  exec "nim c -r " & nimFlagsRefc & " tests/test_serial.nim"
+task test_cbor_serial, "Run CBOR codec unit tests":
+  exec "nim c -r " & nimFlagsOrc & " tests/test_cbor_serial.nim"
+  exec "nim c -r " & nimFlagsRefc & " tests/test_cbor_serial.nim"
```
gmelodie: Idea: maybe we could have CI inject the flags using matrix, so that, at least in CI, `orc` and `refc` tests can run in parallel. Because here refc waits on orc to start.

examples/timer/timer.nim
```diff
@@ -65,19 +65,16 @@ proc nimtimerComplex*(
   return
     ok(ComplexResponse(summary: summary, itemCount: count, hasNote: req.note.isSome))
 
-# --- genBindings() must come AFTER every {.ffi.} / {.ffiCtor.} annotation ---
+proc nimtimer_destroy*(timer: NimTimer) {.ffiDtor.} =
+  ## Tears down the FFI context created by nimtimer_create.
+  ## Blocks until the FFI thread and watchdog thread have joined.
+  discard
+
+# --- genBindings() must come AFTER every {.ffi.} / {.ffiCtor.} / {.ffiDtor.} ---
```
gmelodie: Why is that? Maybe add the reason in the comment as well

examples/nim_timer/rust_bindings/build.rs
```diff
@@ -6,10 +6,6 @@ fn main() {
     let nim_src = manifest.join("../nim_timer.nim");
     let nim_src = nim_src.canonicalize().unwrap_or(manifest.join("../nim_timer.nim"));
 
-    // Walk up to find the nim-ffi repo root (directory containing nim_src's library)
-    // The repo root is where nim c should be run from (contains config.nims).
-    // We assume nim_src lives somewhere under repo_root.
-    // Derive repo_root as the ancestor that contains the .nimble file or config.nims.
```
gmelodie, line 12: Why was this deleted?

ffi/codegen/cpp.nim
```diff
@@ -14,7 +16,7 @@ proc genericInnerType(typeName, prefix: string): string =
 proc nimTypeToCpp*(typeName: string): string =
   let trimmed = typeName.strip()
   if trimmed.startsWith("ptr "):
-    return "void*"
+    return "uint64_t"
```
gmelodie: This is architecture-dependent, if we want to allow 32 bit archs to use ffi, shouldn't this be flexible?

ffi/codegen/cpp.nim
```diff
@@ -32,21 +34,36 @@ proc nimTypeToCpp*(typeName: string): string =
...
   of "float64": "double"
-  of "pointer": "void*"
+  of "pointer": "uint64_t"
   else: trimmed
 
+proc capitalizeFirstLetter(s: string): string =
+  if s.len == 0:
+    return s
+  result = s
+  result[0] = s[0].toUpperAscii()
```
gmelodie: Per the [status Nim style guide](https://status-im.github.io/nim-style-guide/language.result.html) we should avoid using `result` (same for other occurrences)

ffi/codegen/cpp.nim
```diff
@@ -32,21 +34,36 @@ proc nimTypeToCpp*(typeName: string): string =
   of "bool": "bool"
   of "float": "float"
   of "float64": "double"
-  of "pointer": "void*"
+  of "pointer": "uint64_t"
   else: trimmed
 
+proc capitalizeFirstLetter(s: string): string =
```
gmelodie: Maybe use https://nim-lang.org/docs/unicode.html#capitalize%2Cstring from `std/unicode` instead

ffi/codegen/cpp.nim
```diff
@@ -92,10 +109,40 @@ proc generateCppHeader*(
...
+  lines.add("// ============================================================")
+  lines.add("")
+  for p in procs:
+    if p.kind == ffiDtorKind:
+      continue
+    let reqName = reqStructName(p)
+    lines.add("struct $1 {" % [reqName])
+    for ep in p.extraParams:
+      let cppType =
+        if ep.isPtr: "uint64_t"
```
gmelodie: `"uint64_t"` should be a `const` in case we need to change it (e.g. when using 32bit archs) so we can change it in one place only

ffi/codegen/cpp.nim
```diff
@@ -92,10 +109,40 @@ proc generateCppHeader*(
...
+    if fieldNames.len > 0:
       lines.add(
-        "NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE($1, $2)" % [t.name, fieldNames.join(", ")]
+        "NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE($1, $2)" % [reqName, fieldNames.join(", ")]
       )
-      lines.add("")
+    else:
+      # Empty struct round-trips trivially via inline helpers below.
+      lines.add("inline void to_json(nlohmann::json& j, const $1&) { j = nlohmann::json::object(); }" % [reqName])
+      lines.add("inline void from_json(const nlohmann::json&, $1&) {}" % [reqName])
```
gmelodie: I don't get it, why are we dealing with json here?

ffi/codegen/cpp.nim
```diff
@@ -108,62 +155,43 @@ proc generateCppHeader*(
...
+      )
+    of ffiCtorKind:
+      lines.add(
+        "void* $1(const uint8_t* req_cbor, size_t req_cbor_len, FfiCallback callback, void* user_data);" %
+          [p.procName]
+      )
+    of ffiDtorKind:
+      lines.add(
+        "int $1(void* ctx, FfiCallback callback, void* user_data);" % [p.procName]
+      )
```
gmelodie: Would it be better if this were an enum?
```nim
FFIKind = enum
  FFI
  CTOR
  DTOR
```

so you could do 

```suggestion
    case p.kind
    of FFIKind.FFI:
      lines.add(
        "int $1(void* ctx, FfiCallback callback, void* user_data, const uint8_t* req_cbor, size_t req_cbor_len);" %
          [p.procName]
      )
    of FFIKind.CTOR:
      lines.add(
        "void* $1(const uint8_t* req_cbor, size_t req_cbor_len, FfiCallback callback, void* user_data);" %
          [p.procName]
      )
    of FFIKind.DTOR:
      lines.add(
        "int $1(void* ctx, FfiCallback callback, void* user_data);" % [p.procName]
      )
```

ffi/codegen/cpp.nim
```diff
@@ -108,62 +155,43 @@ proc generateCppHeader*(
...
-  # ── Serialization helpers ──────────────────────────────────────────────────
+  # ── CBOR helpers ───────────────────────────────────────────────────────────
   lines.add("template<typename T>")
-  lines.add("inline std::string serializeFfiArg(const T& value) {")
-  lines.add("    return nlohmann::json(value).dump();")
-  lines.add("}")
-  lines.add("")
-  lines.add("inline std::string serializeFfiArg(void* value) {")
-  lines.add("    return std::to_string(reinterpret_cast<uintptr_t>(value));")
+  lines.add("inline std::vector<std::uint8_t> encodeCborFfi(const T& value) {")
```
gmelodie: It would be cool if these function names like `encodeCborFfi` could come from the function declarations themselves in case we need to change them in the future (so we don't have to go grepping everywhere to change.

ffi/codegen/cpp.nim
```diff
@@ -108,62 +155,43 @@ proc generateCppHeader*(
...
+  lines.add("    return nlohmann::json::to_cbor(nlohmann::json(value));")
```
gmelodie: Same as above: even though `to_cbor` comes from an external package, it could change

ffi/codegen/cpp.nim
```diff
@@ -173,31 +201,35 @@ proc generateCppHeader*(
...
+  lines.add("            } else {")
+  lines.add("                s.err.assign(msg, len);")
+  lines.add("            }")
+  lines.add("        }")
   lines.add("        s.done = true;")
   lines.add("        s.cv.notify_one();")
   lines.add("    }")
   lines.add("    delete sptr;")
   lines.add("}")
   lines.add("")
```
gmelodie: This should all come from the function itself. Can we have code that transforms a function into a string? that would reduce errors on the transcriptions and keep it always up to date. We could use the same function to transcribe the function headers to fix my comments from above.

ffi/codegen/cpp.nim
```diff
@@ -228,7 +260,7 @@ proc generateCppHeader*(
   var methods: seq[FFIProcMeta] = @[]
   for p in procs:
     if p.kind == ffiCtorKind: ctors.add(p)
-    else: methods.add(p)
+    elif p.kind == ffiFfiKind: methods.add(p)
```
gmelodie: Nit: `case...of`

ffi/codegen/cpp.nim
```diff
@@ -245,43 +277,53 @@ proc generateCppHeader*(
...
 
-    # -- create() factory --
+    var initList: seq[string] = @[]
+    for n in epNames:
+      initList.add(n)
+    let reqInit =
+      if initList.len > 0:
+        reqName & "{" & initList.join(", ") & "}"
+      else:
+        reqName & "{}"
```
gmelodie: I wonder if we even need this `else`, can we get by with this instead?
```suggestion
      reqName & "{" & initList.join(", ") & "}"
```
The real question: what does the following produce?
```nim
var someList: seq[string] = @[]
someList.join(", ")
```

same for other occurrences

ffi/codegen/cpp.nim
```diff
@@ -32,21 +34,36 @@ proc nimTypeToCpp*(typeName: string): string =
...
+  of "pointer": "uint64_t"
   else: trimmed
 
+proc capitalizeFirstLetter(s: string): string =
+  if s.len == 0:
+    return s
+  result = s
+  result[0] = s[0].toUpperAscii()
+
+proc toCamelCase(s: string): string =
```
fcecin: There's three `toCamelCase` definitions in different files. Better to have just one in some common import.

fcecin (approved):
Epic!
LGTM

ffi/codegen/rust.nim
```diff
@@ -94,10 +97,6 @@ fn main() {
     let nim_src = manifest.join("$1");
     let nim_src = nim_src.canonicalize().unwrap_or(manifest.join("$1"));
 
-    // Walk up to find the nim-ffi repo root (directory containing nim_src's library)
-    // The repo root is where nim c should be run from (contains config.nims).
-    // We assume nim_src lives somewhere under repo_root.
-    // Derive repo_root as the ancestor that contains the .nimble file or config.nims.
```
gmelodie, line 100: Same as before: why was this removed?

ffi/codegen/rust.nim
```diff
@@ -9,23 +9,27 @@ import ./meta
 # ---------------------------------------------------------------------------
 
 proc toSnakeCase*(s: string): string =
-  ## Converts camelCase to snake_case.
-  ## e.g. "delayMs" → "delay_ms", "timerName" → "timer_name"
   result = ""
   for i, c in s:
     if c.isUpperAscii() and i > 0:
       result.add('_')
     result.add(c.toLowerAscii())
```
gmelodie, line 18: 1. Can these string manipulation functions go to their own file and be reused by both cpp.nim and rust.nim? And add tests for them.
2. Names could be more descriptive: `camelToSnakeCase` instead of `toSnakeCase`. 
3. instead of ascii, prefer similar functions from `std/unicode` e.g. `toUpper` or `isUpper` etc.

ffi/codegen/rust.nim
```diff
@@ -211,29 +216,46 @@ proc generateTypesRs*(types: seq[FFITypeMeta]): string =
...
+    let reqName = reqStructName(p)
+    lines.add("#[derive(Debug, Clone, Serialize, Deserialize)]")
+    if p.extraParams.len == 0:
+      lines.add("pub struct $1 {}" % [reqName])
+    else:
+      lines.add("pub struct $1 {" % [reqName])
+      for ep in p.extraParams:
+        let snake = toSnakeCase(ep.name)
+        let rustType =
+          if ep.isPtr: "u64"
```
gmelodie: Same comment as for `cpp`: we should account for other architectures too

ffi/internal/ffi_macro.nim
```diff
@@ -78,65 +71,50 @@ proc bodyHasAwait(n: NimNode): bool =
...
-    result.add newIdentDefs(p[0], p[1])
-
-  when defined(ffiDumpMacros):
-    echo result.repr
+proc nimTypeNameRepr(typ: NimNode): string =
+  ## Stringifies a parameter or field type for the binding-generator registry.
+  ## `$ident` works for simple types; bracket/dot/expression types need `repr`.
+  if typ.kind == nnkIdent: $typ
+  elif typ.kind == nnkPtrTy: "ptr " & nimTypeNameRepr(typ[0])
+  else: typ.repr
```
gmelodie: Nit: use `case..of` instead

ffi/internal/ffi_macro.nim
```diff
@@ -78,65 +71,50 @@ proc bodyHasAwait(n: NimNode): bool =
...
   var procNode = body
   if procNode.kind == nnkStmtList and procNode.len == 1:
     procNode = procNode[0]
   if procNode.kind != nnkLambda and procNode.kind != nnkProcDef:
     error "registerReqFFI expects a lambda proc, found: " & $procNode.kind
 
-  let params = procNode[3] # formal params of the lambda
+  let params = procNode[3]
   var fields: seq[NimNode] = @[]
-  for p in params[1 .. ^1]: # skip return type at index 0
```
gmelodie, line 127: I like it better with the comments

ffi/internal/ffi_macro.nim
```diff
@@ -684,101 +535,105 @@ macro ffi*(prc: untyped): untyped =
...
   let camelName = toCamelCase(procNameStr)
 
-  # Names of generated things
   let reqTypeName = ident(camelName & "Req")
-  let helperProcName = ident(camelName & "Body")
 
-  # Determine whether the body uses async operations
   let isAsync = bodyHasAwait(bodyNode)
 
-  # Strip the * from the exported proc name (needed for both branches)
```
gmelodie, line 694: I like these comments

ffi/internal/ffi_macro.nim
```diff
@@ -684,101 +535,105 @@ macro ffi*(prc: untyped): untyped =
...
+      params = wrapperParams,
+      body = body,
+      pragmas = newTree(nnkPragma, ident("async")),
+    )
+
+  if isAsync:
     # -------------------------------------------------------------------------
-    # 2. registerReqFFI call
+    # ASYNC PATH
     # -------------------------------------------------------------------------
```
gmelodie: This comment can be removed

ffi/cbor_serial.nim
NagyZoltanPeter, line 1: Why not using nim-cbor-serisalization lib?

ffi/internal/ffi_macro.nim
```diff
@@ -856,29 +706,22 @@ macro ffi*(prc: untyped): untyped =
...
-      let retTypeInner = resultInner[1] # RetType from Result[RetType, string]
-      var retIsPtr = false
-      var retTn = ""
-      if retTypeInner.kind == nnkPtrTy:
-        retIsPtr = true
-        retTn = $retTypeInner[0]
-      else:
-        retTn = $retTypeInner
+      let retTypeInner = resultInner[1]
+      let retIsPtr = retTypeInner.kind == nnkPtrTy
```
gmelodie: Maybe extract this in a `isPtr` proc (same for other ocurrences)

ffi/internal/ffi_macro.nim
```diff
@@ -895,41 +738,40 @@ macro ffi*(prc: untyped): untyped =
     result = newStmtList(helperProc, registerReq, ffiProc)
   else:
     # -------------------------------------------------------------------------
-    # SYNC PATH — no await/waitFor in body; bypass thread-channel machinery
+    # SYNC PATH — bypass thread channel; fire callback inline
     # -------------------------------------------------------------------------
```
gmelodie: Instead of these comments, it'd be better to have functions (maybe `syncPath` and `asyncPath`?) that encapsulate the logic

ffi/internal/ffi_macro.nim
```diff
@@ -895,41 +738,40 @@ macro ffi*(prc: untyped): untyped =
...
+        newTree(nnkRecList, syncReqFields)
+      else:
+        newTree(
+          nnkRecList,
+          newTree(nnkIdentDefs, ident("_placeholder"), ident("uint8"), newEmptyNode()),
+        )
+    let syncReqObjTy =
+      newTree(nnkObjectTy, newEmptyNode(), newEmptyNode(), syncReqRecList)
+    let syncReqTypeDef = newNimNode(nnkTypeSection).add(
+      newTree(nnkTypeDef, postfix(reqTypeName, "*"), newEmptyNode(), syncReqObjTy)
```
gmelodie: I have no idea what is going on here, so much so that I can't even suggest improvements, but maybe you could try to make the logic clearer?

ffi/internal/ffi_macro.nim
```diff
@@ -856,29 +706,22 @@ macro ffi*(prc: untyped): untyped =
...
-          isPtr = true
-          tn = $ptype[0]
-        else:
-          tn = $ptype
+        let isPtr = ptype.kind == nnkPtrTy
+        let tn =
+          if isPtr: nimTypeNameRepr(ptype[0])
+          else: nimTypeNameRepr(ptype)
         ffiExtraParams.add(
           FFIParamMeta(name: extraParamNames[i], typeName: tn, isPtr: isPtr)
```
gmelodie: this `isPtr` field seems to never be used, can we remove it?

ffi/internal/ffi_macro.nim
```diff
@@ -1197,27 +1012,33 @@ proc buildCtorProcessFFIRequestProc(
...
-    let ptype = paramTypes[i]
-    newBody.add quote do:
-      let `fieldName` = `reqIdent`[].`fieldName`
-    newBody.add quote do:
-      let `paramName` = ffiDeserialize(`fieldName`, `ptype`).valueOr:
-        return err($error)
+    let paramIdent = ident(paramNames[i])
+    let userType = paramTypes[i]
+    let storedAsString =
+      userType.kind == nnkIdent and $userType == "cstring"
```
gmelodie: This could also be a `proc`, and we could add tests for it

ffi/internal/ffi_macro.nim
```diff
@@ -1492,33 +1288,19 @@ macro ffiCtor*(prc: untyped): untyped =
...
+    typeDef, ffiNewReqProc, helperProc, processProc, addToReg, poolDecl,
     ffiProc,
   )
 
   when defined(ffiDumpMacros):
     echo result.repr
 
 # ---------------------------------------------------------------------------
-# ffiDtor — destructor macro
+# ffiDtor — destructor macro (unchanged signature)
```
gmelodie:
```suggestion
# ffiDtor — destructor macro
```

tests/test_gc_compat.nim
```diff
@@ -143,10 +154,12 @@ suite "GC safety - string lifetime across thread boundary":
...
     check d.retCode == RET_OK
-    check d.msgLen == 512
-    check d.msg[0] == 'a'
-    check d.msg[25] == 'z'
-    check d.msg[26] == 'a'
+    # CBOR-encoded 512-char text string: 3 header bytes (0x79 + u16 len) + 512 chars.
+    let decoded = callbackOkString(d)
+    check decoded.len == 512
+    check decoded[0] == 'a'
+    check decoded[25] == 'z'
```
gmelodie: For encoding tests, instead we could do
- take an obj
- encode the obj
- decode the obj

this asserts encoding and decoding are compatible and consistent

ffi/ffi_thread_request.nim
```diff
@@ -1,69 +1,105 @@
...
   ret[].callback = callback
   ret[].userData = userData
   ret[].reqId = reqId.alloc()
-  ret[].reqContent = reqContent
+  ret[].dataLen = data.len
+  if data.len > 0:
+    ret[].data = cast[ptr UncheckedArray[byte]](allocShared(data.len))
+    copyMem(ret[].data, unsafeAddr data[0], data.len)
+  else:
+    ret[].data = nil
```
gmelodie:
```suggestion
  ret[].data = nil
  if data.len > 0:
    ret[].data = cast[ptr UncheckedArray[byte]](allocShared(data.len))
    copyMem(ret[].data, unsafeAddr data[0], data.len)
```
same for other occurrences

gmelodie (commented):
Overall coments
- nit: `NimTimer*` -> `Timer*`
- Remove `result` usage
- Don't hardcode function names in the strings in `fii/codegen`.
- I think this code would be more readable if the macros were define hierarchically, like the language's grammar: e.g. in rust the starting point would always be `fromRust` or `toRust`, and expressions would be expanded from there.

Note: these long PRs make it harder to understand the architecture of what you're looking for. Smaller PRs (~<400 lines of code) are better to understand and yield more productive reviews

tests/test_cbor_serial.nim
```diff
@@ -0,0 +1,166 @@
...
+    let bytes = cborEncode(s)
+    check cborDecode(bytes, string).value == s
+
+  test "string with special chars":
+    let s = "tab\there\nnewline"
+    let bytes = cborEncode(s)
+    check cborDecode(bytes, string).value == s
+
+suite "CBOR pointer / ptr":
+  test "pointer round-trip":
```
NagyZoltanPeter: Why allow pointer transmit on interface level? It's a big security risk, because GC other party over FFI gate will have no information about memory state.

examples/timer/timer.nim
```diff
@@ -2,12 +2,12 @@ import ffi, chronos, options
...
 
-declareLibrary("nimtimer")
-
 # The library's main state type. The FFI context owns one instance.
 type NimTimer = object
   name: string # set at creation time, read back in each response
 
+declareLibrary("nimtimer", NimTimer)
+
 type TimerConfig {.ffi.} = object
```
NagyZoltanPeter, line 11: Sorry cannot get from generator code if we are gonna force API interface to use composed objects in case of multiple argument for an API request interface?
Or is it possible to do it natural like `send(content_topic, payload, ephemeral)`? Maybe next version?

ffi/codegen/cpp.nim
Ivansete-status: Addressed in 95c5097bb12e9af84c35a784f510d4bcae26b159

NagyZoltanPeter (changes requested):
I found it better to discuss this review through with Claude... and now with you @Ivansete-status .

Please read the outcome of my intentional driven review with AI.

# Findings — `start-using-cbor` PR review

## Sync fast-path (`{.ffi.}` without `await`)

**Trigger**: `ffi/internal/ffi_macro.nim:540` — `let isAsync = bodyHasAwait(bodyNode)` → if false, takes branch at `ffi_macro.nim:740`.

**Effect**: C export runs CBOR decode + user body + CBOR encode **inline on the foreign caller thread**, bypassing the FFI thread, the channel, `ctx.lock`, and `foreignThreadGc`.

| # | Finding | Where | Severity | Why |
|---|---|---|---|---|
| 1 | No `foreignThreadGc` wrap on sync path | `ffi_macro.nim:776-820` | 🟥 **Critical (refc)** | Foreign thread has no per-thread heap. First Nim allocation (CBOR decode at line 791, or any `string`/`seq` in user body) → crash or silent heap corruption. Hits every refc build. |
| 2 | Concurrent access to `ctx.myLib` from foreign thread vs. FFI thread | `ffi_macro.nim:816-820` (`ctx.myLib[]` deref on caller thread) | 🟥 **Critical (all backends)** | Sync handler runs while async handler may be mid-execution. Non-atomic refcount on `ref`/`seq`/`string` → torn reads, double-free, UAF under ORC. UB under refc. Not fixable by user code review. |
| 3 | chronos primitives reachable from foreign thread | same — anything in lib state | 🟥 **Critical** | chronos dispatcher is single-threaded by contract. Touching a `Future`/`AsyncLock`/`AsyncEvent` from the foreign thread is UB even just reading. |
| 4 | Branch selection by AST heuristic (`bodyHasAwait`) | `ffi_macro.nim:64-67, 540` | 🟧 **High** | User has no source-level signal which thread their handler runs on. Removing/adding an `await` silently relocates execution and changes memory model. Refactor hazard. |
| 5 | Declared `Future[Result[T, string]]` lies about threading | `ffi_macro.nim:588-605` (wrapper) | 🟧 **High** | Signature is async, execution is not. Misleads reviewers and the lib author about the threading contract. |
| 6 | No reentrancy guard on `sendRequestToFFIThread` | `ffi/ffi_context.nim` — `sendRequestToFFIThread` | 🟧 **High** | Handler running on FFI thread that calls an async `{.ffi.}` symbol on itself → self-deadlock (waits on `reqReceivedSignal` it would itself fire). Cheap to detect; currently undetected. |
| 7 | `ctx.lock` held across `trySend` + `fireSync` + `waitSync` | `ffi/ffi_context.nim` — `sendRequestToFFIThread` | 🟨 **Medium** | All async submissions serialise on this lock; multiple `xxxAsync` callers from C++ pile up at the mutex before reaching Nim. Known limitation (comment present), but worth re-stating with the CBOR rework. |
| 8 | C++ `xxxAsync` wrappers misleading | `examples/nim_timer/cpp_bindings/nimtimer.hpp` (`std::async(std::launch::async, ...)`) | 🟨 **Medium** | Implies parallelism that the single FFI thread + serialising lock cannot deliver. Spawns thread per call (libc++/libstdc++ behavior), timeout clock drifts. Should be removed or renamed. |
| 9 | `cborEncode` allocates GC `seq[byte]` then `copyMem` into shared heap | `ffi/ffi_thread_request.nim:36-37` | 🟨 **Medium** | Double allocation per request. Add a `cborEncodeShared` later if request payloads grow. |
| 10 | No test for `trySend`-failure rollback with new ownership | `ffi/ffi_context.nim` — `sendRequestToFFIThread` | 🟩 **Low** | Path exists but uncovered. Add a fault-injection test. |
| 11 | `CborNullByte` sentinel via `addr` of a stack local | `ffi/ffi_thread_request.nim:99-101` | 🟩 **Low** | Safe today (callback is synchronous), fragile if async callback path is ever introduced. Add a comment. |

## Severity legend

| | Meaning | Action |
|---|---|---|
| 🟥 **Critical** | Memory unsafety / undefined behavior under realistic use. Production-blocker. | Must fix before merge. |
| 🟧 **High** | Correctness footgun the framework cannot prevent and the user cannot audit. | Fix or document with explicit opt-in pragma. |
| 🟨 **Medium** | Performance / API ergonomics / undertested path. Won't corrupt memory but misleads users or wastes resources. | Fix before next release. |
| 🟩 **Low** | Defensive / hygiene. | Address opportunistically. |

## Recommended single fix that closes findings 1–5

**Delete the `if isAsync:` branch at `ffi_macro.nim:607`; always emit the async path.**

- **Cost**: ~one channel hop + one signal per call (microseconds).
- **Benefit**: one threading model, one memory model, no inference from `await` presence, all `{.ffi.}` procs verifiably safe under both refc and ORC.

If a real fast-path is wanted later, introduce it as an explicit `{.ffiSync.}` pragma with a documented contract ("must not allocate GC memory, must not touch lib state"). Never infer threading from body contents.

examples/nim_timer/cpp_bindings/nimtimer.hpp
Ivansete-status: Addressed in e19f29aef85d233b7e792efa03f819e7537d0d77

NagyZoltanPeter:
Rust binding extension critics and proposal by Claude - as I'm not an expert in rust...

# Rust bindings review — `start-using-cbor`

**Overall**: Better than the C++ bindings (async path uses tokio idiomatically), but the blocking trampoline has structural correctness bugs and the context type is missing essential lifecycle management. **Not production-shippable as generated.**

## Findings

| # | Finding | Where | Severity |
|---|---|---|---|
| 1 | `Arc::from_raw` + `mem::forget` in `on_result` leaks every successful call (count never decremented to zero) | `api.rs` blocking trampoline | 🟥 **Critical** |
| 2 | Timeout path leaks the user-data allocation permanently; late callback after timeout writes into freed memory | `ffi_call` after `wait_timeout_while` | 🟥 **Critical** |
| 3 | **No `Drop` impl** on `NimTimerCtx` — forgetting the ctx leaks the entire Nim runtime (FFI thread, watchdog, chronos, lib state) | `api.rs` ctx struct | 🟥 **Critical** |
| 4 | Async path (`ffi_call_async`) has **no timeout** — hangs forever if Nim never fires the callback | `ffi_call_async` | 🟧 **High** |
| 5 | `new_async` ignores caller-provided timeout, hardcodes 30s | `NimTimerCtx::new_async` | 🟧 **High** |
| 6 | Cancellation does not reach Nim — dropping the future just discards the eventual response, broker keeps running | both `*_async` methods | 🟧 **High** |
| 7 | `unsafe impl Send for NimTimerCtx` / `unsafe impl Sync for NimTimerCtx` blanket-applied with no safety comment or justification | `api.rs` after ctx struct | 🟧 **High** |
| 8 | Blocking path uses `std::sync::Mutex + Condvar`; async path uses `tokio::oneshot`. Two separate trampolines for the same purpose | `ffi_call` vs `ffi_call_async` | 🟨 **Medium** |
| 9 | `ffi_call` clones the payload `Vec<u8>` from the locked state instead of `mem::take` — extra allocation per call | end of `ffi_call` | 🟨 **Medium** |
| 10 | `String::from_utf8_lossy` silently rewrites non-UTF-8 error bytes | `on_result`, `on_result_async` | 🟨 **Medium** |
| 11 | `tokio` is a hard dependency even for users who only use the blocking API | `Cargo.toml` | 🟨 **Medium** |
| 12 | `#[link(name = "nimtimer")]` with no static/dynamic switch, no `build.rs` knob | `ffi.rs` | 🟨 **Medium** |
| 13 | Errors are stringly-typed (`Result<_, String>`); no error enum, breaks `?` ergonomics for downstream callers | `api.rs` throughout | 🟩 **Low** |
| 14 | All generated types/helpers `pub` at crate root; wide unstable surface | `api.rs`, `types.rs`, `ffi.rs` | 🟩 **Low** |
| 15 | Generated `Deserialize` types lack `deny_unknown_fields` — silently accepts Nim-side schema drift | `types.rs` | 🟩 **Low** |

## Severity legend

| | Meaning |
|---|---|
| 🟥 **Critical** | Memory unsafety, resource leak on every call, or missing required lifecycle. Production-blocker. |
| 🟧 **High** | Correctness gap or API misuse that the user cannot detect without reading the generated source. |
| 🟨 **Medium** | Performance or ergonomics flaw; works but wastes resources or misleads users. |
| 🟩 **Low** | Polish / idiomatic-Rust hygiene. |

## Root cause

The blocking and async trampolines were designed independently and use **two different ownership models**:

- **Async** (`oneshot::Sender` in a `Box`): one allocation, callback consumes it once, drops cleanly. **Almost correct** — only missing timeout + cancellation propagation.
- **Blocking** (`Arc<(Mutex, Condvar)>` + `Arc::from_raw` + `mem::forget`): the `Arc` count math is wrong; **every call leaks**, and the timeout path is unsound.

The split also forces the user to choose `tokio` even if they only want blocking calls.

## Proposed fix — replace both trampolines with one

Use **one** primitive that works in both worlds: a `flume::bounded(1)` channel (or any oneshot supporting both `recv_timeout` and async `recv_async`).

```rust
fn ffi_call(timeout: Duration, f: impl FnOnce(...)) -> Result<Vec<u8>, String> {
    let (tx, rx) = flume::bounded::<Result<Vec<u8>, String>>(1);
    let raw = Box::into_raw(Box::new(tx)) as *mut c_void;
    let ret = f(on_result, raw);
    if ret == RET_MISSING_CALLBACK {
        drop(unsafe { Box::from_raw(raw as *mut flume::Sender<_>) });
        return Err("missing callback".into());
    }
    rx.recv_timeout(timeout).map_err(|_| "timeout".into())?
}

async fn ffi_call_async(timeout: Duration, f: impl FnOnce(...)) -> Result<Vec<u8>, String> {
    // identical shape, uses rx.recv_async() with tokio::time::timeout
}

unsafe extern "C" fn on_result(ret, msg, len, ud) {
    let tx = Box::from_raw(ud as *mut flume::Sender<Result<Vec<u8>, String>>);
    let _ = tx.send(parse(ret, msg, len));  // drops on scope exit — single ownership
}
```
Properties this gives you:

One allocation per call, owned exclusively by the C side until the callback fires.
No mem::forget, no reference-count gymnastics, no leak.
Timeout-safe on both paths — async via tokio::time::timeout, sync via recv_timeout. Late callbacks after timeout drop into a closed receiver; no UAF, no leak.
Single trampoline — async and blocking diverge only at the recv call. Cuts ~80 lines of generated Rust.
No tokio in the dependency tree for blocking-only users — gate the async API behind a cargo feature.

## Additional must-have changes

| Change | Why |
|---|---|
| **Add `Drop` for `*Ctx`** that calls `<lib>_destroy(self.ptr, ...)` | Without it, leaking a ctx leaks the Nim runtime. The C++ binding already does this. |
| **Async ctors and methods must take `Duration`** consistent with sync versions | `new_async` currently hardcodes 30s, `*_async` methods have no timeout at all. |
| **Document `unsafe impl Send` / `Sync`** with a safety comment referencing the Nim-side serialisation guarantees | Currently silent; trips `clippy::missing_safety_doc` and is a footgun if Nim threading changes. |
| **Replace `String` errors with a `thiserror` enum** (`FfiError::Timeout`, `FfiError::Cbor`, `FfiError::Nim(String)`, `FfiError::Internal`) | Idiomatic Rust; downstream `?` works cleanly. |
| **Feature-gate the async API**: `[features] async = ["dep:tokio"]`, blocking is default | Don't force `tokio` on consumers who don't need it. |
| **Provide a `build.rs`** that handles static vs dynamic linking and locates the Nim-built artifact | Match what the C++ CMakeLists.txt does. |
| **Wrap raw FFI in `mod ffi { ... }`** and re-export only the high-level types | Narrows the public surface and signals what's stable. |

## Cancellation note

Drop-cancels-the-future only works if Nim cooperates. Without a cancellation token in the C ABI, the broker keeps running the dropped request to completion and discards the result. This is the **same problem `xxxAsync` has in C++** and the same reason async FFI wrappers are partially aspirational under a single-threaded broker. **Not fixable in the binding alone** — needs:

1. A `cancel_token: *const AtomicBool` (or equivalent) in the C call signature, **or**
2. A separate `<lib>_cancel(ctx, request_id)` entry point and request IDs.

Worth raising as a follow-up; cancellation is a property of the system, not the binding.

## TL;DR

- **3 critical bugs** to fix before merge: leaked `Arc` (every call), unsound timeout path, missing `Drop` on context.
- **3 high-priority gaps**: async timeout missing, async ctor ignores timeout, no cancellation propagation.
- **Proposal**: replace the two trampolines with a single `flume`-based one (~80 lines deleted, leak + timeout bugs gone in one move, blocking-only users no longer need `tokio`).
- **Cancellation** needs C-ABI support to work properly — track separately.
- Compared to C++: design intent is more idiomatic, execution has more bugs. The path to "good Rust bindings" is shorter than the path to "good C++ bindings" — most fixes are one-liners once the unified trampoline is in.

examples/nim_timer/cpp_bindings/nimtimer.hpp
Ivansete-status: Addressed in 6b27d896a2e277ee4da3faf7956162a45ec251a5

examples/nim_timer/cpp_bindings/nimtimer.hpp
Ivansete-status: Addressed in ffe73e8c05d0bab2b3c8ca43f0addd454736d926

examples/timer/rust_bindings/src/api.rs
Ivansete-status: Addressed in 3660b2a12854ad779502b1a213b83e6ae95865ff

examples/timer/rust_bindings/src/api.rs
Ivansete-status: I've added comments to document better this `fn` in the following: ebb439ee9481421171e7db45679e392a46a24a71

examples/timer/rust_bindings/src/api.rs
Ivansete-status: In this particular case I'm afraid this is the idiomatic manner to put the sender channel side in the heap and make it available for another thread. For `rx` there's no need to box it. We could simplify the `fn` a bit by using alias types. I can add a comment to explain better the purpose of that if you want.

ffi.nimble
Ivansete-status: Good one! We will tackle that in: https://github.com/logos-messaging/nim-ffi/issues/25

examples/timer/timer.nim
Ivansete-status: Changed the wording in the following: 566ffba1ad4cd01b65a447283c24700eb96cc5be

examples/nim_timer/rust_bindings/build.rs
Ivansete-status: Reverted in e89fd9915e32be62fb8ac8d7fd81e6029ef44583

ffi/codegen/cpp.nim
Ivansete-status: Yes good point! This is the answer I got from Claude to the question "can it cause issues in a 32b arch?"
image

ffi/codegen/cpp.nim
Ivansete-status: Addressed in 19a4539620760dda04d7fa24c194a7f2f6e3b7c1

ffi/codegen/cpp.nim
Ivansete-status: That proc has been removed in 95c5097bb12e9af84c35a784f510d4bcae26b159

ffi/codegen/cpp.nim
Ivansete-status: Addressed in fc50639bd7386cb3a9f8d7cc77a435986e8526e3

ffi/codegen/cpp.nim
Ivansete-status: This is a `nlohmann` thing, which needs to use json before cbor.
struct -> json -> cbor

ffi/codegen/cpp.nim
Ivansete-status: Addressed in 5e112783c6f2d3518b3233741c6d83b0639a1e2f

ffi/codegen/cpp.nim
Ivansete-status: Sorry, I can't quite follow this one. May share an example :) ?

ffi/cbor_serial.nim
Ivansete-status: To avoid another dependency. But we can add it if required.

NagyZoltanPeter: Why that dependency bother? I cannot tell if this hand rolled implementation is aligned with that serialization lib or not. Seems to me a double validation need due code dup.

ffi/codegen/cpp.nim
Ivansete-status: I think the following addressed that 23fcbfa534d85c496d70e1c219f58045fb02826e

ffi/codegen/cpp.nim
Ivansete-status: Addressed in 86c46da307eb8ada17232fc41297ae2bad7635c3

ffi/codegen/cpp.nim
Ivansete-status: Addressed in 16b22aee6868209049849eba710f8a5bae33d4ff

ffi/codegen/rust.nim
Ivansete-status: Addressed in 4ddeb13c7ec2d75abf02692c2dd8b9a314f8c9ad

ffi/codegen/rust.nim
Ivansete-status: Addressed in 881cc818dd9505e07dee4c046e797b7afc397cf6

ffi/codegen/rust.nim
Ivansete-status: Created a const in case we want to adapt it in the upcoming future 3726bbc13580358f6660aecf5e1f2050bd92533a

ffi/internal/ffi_macro.nim
Ivansete-status: Addressed in 08c6b523bdb4164c9dc816cdb04f943a13661ef1

ffi/internal/ffi_macro.nim
Ivansete-status: Addressed in 627eb77adde582e7e809b4417cf2f9da40f8c3b4

ffi/internal/ffi_macro.nim
Ivansete-status: Addressed in 627eb77adde582e7e809b4417cf2f9da40f8c3b4

ffi/internal/ffi_macro.nim
Ivansete-status: Addressed in 627eb77adde582e7e809b4417cf2f9da40f8c3b4

ffi/internal/ffi_macro.nim
Ivansete-status: Addressed in 745c251567ea7a3922e869fb64b7d24a7fb25bcb

ffi/internal/ffi_macro.nim
Ivansete-status: Addressed in 0be88164fe79133fda5815f0d7afe60f7ffcdc40

ffi/internal/ffi_macro.nim
Ivansete-status: Well, when dealing with macros we have to deal with this weird code.
The best way to understand it is by passing -d:ffiDumpMacros compile flag, in which case, at compile time, you can see the generated procs.

On the other hand, in order to generate a macro definition from Nim snippet, `dumpAstGen` can be used. For example:
```nim
dumpAstGen:
  proc hello(x: int): int =
    x + 1
```

So, dealing with macros is a tricky work that needs to be done combining `type.repr`, `dumpAstGen`, and AI. And the macro code may always be very difficult to read directly. 

Addressed in b4b66e37442d731a7907d6488b6801c2c620d00c

ffi/internal/ffi_macro.nim
Ivansete-status: `isPtr` is needed at codegen. Added this to add some more clarity: 059fd649260d0169ce774a9ef79353ba44fdc127

ffi/internal/ffi_macro.nim
Ivansete-status: Addressed in b98b43a612322481d1503cf6cf7afe6469927959

ffi/internal/ffi_macro.nim
Ivansete-status: Cannot commit directly as diff is outdated. Addressed in 1f9361f54b4b903eedc0b2479dcb4815d21fd5a8

tests/test_gc_compat.nim
Ivansete-status: Addressed in 292f01ab5f4c206f067c1a4358b4e81bfc1403b1

ffi/ffi_thread_request.nim
Ivansete-status: Addressed in 8ecf6f85a1785e2bfc48c4daaabf187f93f156d8

Ivansete-status:
The recent commits tackle 1-5, 6, 7, 9 issues from https://github.com/logos-messaging/nim-ffi/pull/23#pullrequestreview-4283811768

Ivansete-status:
The previous commits cover the critical points for Rust comment: https://github.com/logos-messaging/nim-ffi/pull/23#issuecomment-4444807463

NagyZoltanPeter (approved):
I think as the main concerns gone it can go.
But I would extend the tests coverage hugely. Need for e2e tests in C/C++/Rust exercising full circle. Stress tests are also essential to reveal threading weaknesses than only come under heavy load.

ffi/codegen/templates/cpp/context_rule_of_5.hpp.tpl
```diff
@@ -0,0 +1,22 @@
+    ~{{CTX}}() {
```
gmelodie, line 13: Might be worth it to add a comment explaining what `rule_of_5` means (idk what it is tbh)

ffi/cbor_serial.nim
gmelodie: Agreed with @NagyZoltanPeter, if other people are using that other lib, we should prefer it for compat sake

ffi/ffi_thread_request.nim
```diff
@@ -1,69 +1,135 @@
...
+  ret[].reqId = reqId.alloc()
+  ret[].data = nil
+  ret[].dataLen = 0
+  if dataLen > 0 and not data.isNil():
+    ret[].data = data
+    ret[].dataLen = dataLen
+  elif not data.isNil():
+    # `cborEncodeShared` returns `(nil, 0)` for empty payloads, but be safe
+    # if a caller hands us a zero-length-but-non-nil buffer.
+    deallocShared(data)
```
gmelodie: nit: this could be encapsulated in a proc

```suggestion
  ret[].data = copyData(data, dataLen)
```

ffi/codegen/meta.nim
```diff
@@ -44,3 +45,39 @@ const ffiOutputDir* {.strdefine.} = ""
 # Nim source path (relative to outputDir) embedded in generated build files;
 # set with -d:ffiNimSrcRelPath=../relative/path.nim
 const ffiNimSrcRelPath* {.strdefine.} = ""
+
+# ---------------------------------------------------------------------------
+# Name conversion helpers shared by codegen and the ffi macro
+# ---------------------------------------------------------------------------
+
+proc camelToSnakeCase*(s: string): string =
```
gmelodie: nit: I'd like it better if these string manipulations were inside a `utils.nim`, (decoupled from `meta.nim`) and their tests as well (`test_utils.nim`)

tests/test_gc_compat.nim
```diff
@@ -50,9 +50,21 @@ proc waitCallback(d: var CallbackData) =
...
+proc callbackBytes(d: var CallbackData): seq[byte] =
+  var bytes = newSeq[byte](d.msgLen)
+  if d.msgLen > 0:
+    copyMem(addr bytes[0], addr d.msg[0], d.msgLen)
+  return bytes
+
+proc callbackOkString(d: var CallbackData): string =
+  ## Decodes the CBOR success payload as a string.
+  cborDecode(callbackBytes(d), string).valueOr:
+    return ""
```
gmelodie, line 73: We should re-check that `RET_OK` == `true` instead of assuming (maybe use `doAssert` for this)

ffi/internal/ffi_macro.nim
```diff
@@ -13,15 +13,18 @@ when defined(ffiGenBindings):
...
   ## Leaves already-snake_case names unchanged.
-  ## e.g. "nimtimerCreate" → "nimtimer_create", "nimtimer_echo" → "nimtimer_echo"
+  ## e.g. "timerCreate" → "timer_create", "timer_echo" → "timer_echo"
+  var snake = ""
   for i, c in s:
     if c.isUpperAscii() and i > 0:
-      result.add('_')
-    result.add(c.toLowerAscii())
+      snake.add('_')
+    snake.add(c.toLowerAscii())
```
gmelodie: unicode-aware conversions (same for other occurrences)
```suggestion
    snake.add(c.toLower())
```

ffi/internal/ffi_macro.nim
```diff
@@ -13,15 +13,18 @@ when defined(ffiGenBindings):
 proc nimNameToCExport(s: string): string =
   ## Converts a camelCase Nim proc name to a snake_case C export name.
   ## Leaves already-snake_case names unchanged.
-  ## e.g. "nimtimerCreate" → "nimtimer_create", "nimtimer_echo" → "nimtimer_echo"
+  ## e.g. "timerCreate" → "timer_create", "timer_echo" → "timer_echo"
+  var snake = ""
   for i, c in s:
     if c.isUpperAscii() and i > 0:
```
gmelodie: unicode-aware conversions (same for other occurrences)

```suggestion
    if c.isUpper() and i > 0:
```

ffi/internal/ffi_macro.nim
```diff
@@ -42,259 +45,244 @@ proc registerFfiTypeInfo(typeDef: NimNode): NimNode {.compileTime.} =
...
+
+proc fieldStorageType(typ: NimNode): NimNode =
+  ## Returns the in-Req-struct storage type for a user-declared param type.
+  ## `cstring` is stored as `string` for trivial CBOR transport; everything
+  ## else is stored as the user typed it.
+  case typ.kind
+  of nnkIdent:
+    if $typ == "cstring": ident("string")
+    else: typ
+  else: typ
```
gmelodie: don't remember if it was my idea to use `case...of` here, but apologies if so, this way is better
```suggestion
  if typ.kind == nnkIdent and $typ == "cstring":
    return ident("string")
  return typ
```

ffi/internal/ffi_macro.nim
```diff
@@ -42,259 +45,244 @@ proc registerFfiTypeInfo(typeDef: NimNode): NimNode {.compileTime.} =
...
+
+proc nimTypeNameRepr(typ: NimNode): string =
+  ## Stringifies a parameter or field type for the binding-generator registry.
+  ## `$ident` works for simple types; bracket/dot/expression types need `repr`.
+  case typ.kind
+  of nnkIdent: $typ
+  of nnkPtrTy: "ptr " & nimTypeNameRepr(typ[0])
+  else: typ.repr
+
+proc fieldStorageType(typ: NimNode): NimNode =
```
gmelodie: nit
```suggestion
proc storageType(typ: NimNode): NimNode =
```

ffi/internal/ffi_macro.nim
```diff
@@ -42,259 +45,244 @@ proc registerFfiTypeInfo(typeDef: NimNode): NimNode {.compileTime.} =
...
+  ##
+  ## Built with the runtime AST API rather than `quote do:` so the proc is
+  ## callable from both macro context and ordinary code (e.g. unit tests).
+  let storedAsString = userType.kind == nnkIdent and $userType == "cstring"
+  if storedAsString:
+    let fieldAccess = newDotExpr(decodedIdent, fieldIdent)
+    let castExpr = newDotExpr(fieldAccess, ident("cstring"))
+    return nnkLetSection.newTree(
+      nnkIdentDefs.newTree(fieldIdent, ident("cstring"), castExpr)
+    )
```
gmelodie: Invert the logic here

```nim
if not storedAsString:
  return ...
let fieldAccess...
return nnkLetSection...
```

ffi/internal/ffi_macro.nim
```diff
@@ -42,259 +45,244 @@ proc registerFfiTypeInfo(typeDef: NimNode): NimNode {.compileTime.} =
...
-  ##     ): Future[Result[string, string]] {.async.} =
-  ##       ...
-  ## The extracted fields will be:
-  ##   - configJson: cstring
-  ##   - appCallbacks: AppCallbacks
-  ## 
+    return newLetStmt(fieldIdent, newDotExpr(decodedIdent, fieldIdent))
+
+proc cExportedParams(ctxType: NimNode): seq[NimNode] =
+  ## Standard parameter list for the C-exported wrapper of a .ffi. proc:
```
gmelodie, line 120:
```suggestion
  ## Standard parameter list for the C-exported wrapper of an .ffi. proc:
```

ffi/internal/ffi_macro.nim
```diff
@@ -856,238 +791,77 @@ macro ffi*(prc: untyped): untyped =
...
-        var tn = ""
-        if ptype.kind == nnkPtrTy:
-          isPtr = true
-          tn = $ptype[0]
-        else:
-          tn = $ptype
+        let isPointer = isPtr(ptype)
+        let tn =
+          if isPointer: nimTypeNameRepr(ptype[0])
+          else: nimTypeNameRepr(ptype)
```
gmelodie: can you not do this instead?
```suggestion
        let tn =
          if isPtr(ptype): nimTypeNameRepr(ptype[0])
          else: nimTypeNameRepr(ptype)
```

ffi/internal/ffi_macro.nim
```diff
@@ -856,238 +791,77 @@ macro ffi*(prc: untyped): untyped =
...
-      if retTypeInner.kind == nnkPtrTy:
-        retIsPtr = true
-        retTn = $retTypeInner[0]
-      else:
-        retTn = $retTypeInner
+      let retTypeInner = resultInner[1]
+      let retIsPtr = isPtr(retTypeInner)
+      let retTn =
+        if retIsPtr: nimTypeNameRepr(retTypeInner[0])
+        else: nimTypeNameRepr(retTypeInner)
```
gmelodie: Same as above

ffi/internal/ffi_macro.nim
```diff
@@ -1282,146 +1044,143 @@ proc addCtorRequestToRegistry(reqTypeName, libTypeName: NimNode): NimNode =
...
-  let helperProcNameStr = reqTypeNameStr[0 ..^ ("CtorReq".len + 1)] & "CtorBody"
-  let helperProcName = ident(helperProcNameStr)
+  # The user-facing Nim proc keeps the user's original name with their declared
+  # signature; the C-exported wrapper moves to `<userProcName>ExportC` and
+  # binds the snake_case C symbol via `{.exportc.}`.
+  let userProcName =
+    if procName.kind == nnkPostfix:
+      procName[1]
+    else:
+      procName
```
gmelodie: nit
```suggestion
  var userProcName = procName
  
  if procName.kind == nnkPostfix:
    userProcName = procName[1]
```

ffi/internal/ffi_macro.nim
```diff
@@ -1461,51 +1213,47 @@ macro ffiCtor*(prc: untyped): untyped =
...
-        isPtr = true
-        tn = $ptype[0]
-      else:
-        tn = $ptype
-      ctorExtraParams.add(FFIParamMeta(name: paramNames[i], typeName: tn, isPtr: isPtr))
+      let isPointer = isPtr(ptype)
+      let tn =
+        if isPointer: nimTypeNameRepr(ptype[0])
+        else: nimTypeNameRepr(ptype)
+      ctorExtraParams.add(FFIParamMeta(name: paramNames[i], typeName: tn, isPtr: isPointer))
```
gmelodie: Same as above

ffi/codegen/cpp.nim
gmelodie: wtf

gmelodie (commented):
Only one major comment that I'd really like to see before approving:
- Agreed with @NagyZoltanPeter about not cbor-serializing it ourselves if there's a nim lib that does it. (I didn't know that existed)

Nit: I like `FFI` better than `Ffi`, the latter is hard to read as FFI

ffi/cbor_serial.nim
Ivansete-status: Yes, fair enough. Addressed in 9602b5d81b9f3ab586d00607d0f3b9033ec8513b

examples/timer/timer.nim
Ivansete-status: Yes good point, thanks! Addressed in bdb6498c6d20c2686f91b95138527a7bbeebda1f

gmelodie (approved):
FUCK YEA

tests/test_cbor_serial.nim
Ivansete-status: Yes thanks for highlighting it. More protected in at compile time 6275e6ff6a3bc7a7562a48fb92f896bb05131e72

ffi/codegen/templates/cpp/context_rule_of_5.hpp.tpl
Ivansete-status: Addressed in 4dffdaa86c22384a93e95cf27196d8f9d3f2d258

ffi/ffi_thread_request.nim
Ivansete-status: Addressed in a78297e3d087712990a8177c9239cf43a7b3f696

ffi/codegen/meta.nim
Ivansete-status: Addressed in ad49ab51c2f7f7718c8ee63c8bde8fd21fc57042

tests/test_gc_compat.nim
Ivansete-status: Addressed in 0214c9044103f8c0fc347f10060841bf2362d77f

ffi/internal/ffi_macro.nim
Ivansete-status: Addressed in 25a538cc568c636c5b2954e4a20c3a7cb7d8b6a9

ffi/internal/ffi_macro.nim
Ivansete-status: Also addressed in 25a538cc568c636c5b2954e4a20c3a7cb7d8b6a9

ffi/internal/ffi_macro.nim
Ivansete-status: Addressed in aceaa2f2b2ccfc95aa6f8aaba02e4996592319d9

ffi/internal/ffi_macro.nim
Ivansete-status: Addressed in ce55059c9bb70b7863a2a149cabf3d1c2cc94389

ffi/internal/ffi_macro.nim
Ivansete-status: Addressed in 4150750900c169f15b83674cab56da52d9744938

ffi/internal/ffi_macro.nim
Ivansete-status: I'm afraid not possible as `isPointer` is used a line below

ffi/internal/ffi_macro.nim
Ivansete-status: Addressed in 4d6c3a64169fe9ef0ee6f1e4ff1124c658a17260

ffi/internal/ffi_macro.nim
Ivansete-status: Addressed in 4d6c3a64169fe9ef0ee6f1e4ff1124c658a17260

ffi/codegen/cpp.nim
Ivansete-status: Addressed in the following: 4da34084d4e34123a1d597c399d1afd76274f4ac
Use [TinyCBOR](https://github.com/intel/tinycbor) instead.

PR #28, opened 2026-05-18
Add aggregator and stress tests

Ivansete-status:


Ivansete-status:
Closing as not explicit need. We will cover san issue that will give great tests.

PR #29, opened 2026-05-18
Add raw ffi mode

Ivansete-status:
## Summary

Replaces the single-axis `-d:targetLang=` knob with a 2-axis matrix:

-d:ffiMode={ raw | cbor }    # HOW data crosses the FFI boundary
-d:ffiLang={ cpp | rust }    # WHICH language the consumer wrapper uses

|                  | `ffiLang=cpp`                                                    | `ffiLang=rust`                              |
|------------------|------------------------------------------------------------------|----------------------------------------------|
| `ffiMode=cbor`   | C++ header, CBOR on the wire — in-process **and** inter-process | Rust crate, CBOR on the wire — both shapes  |
| `ffiMode=raw`    | Flat C ABI header — in-process only, zero-copy-friendly         | 🚧 not yet implemented                       |

## Motivation

For high-throughput in-process consumers (e.g. status-app delivering Waku messages up to 150 KiB at hundreds of msg/s), CBOR encode/decode is a meaningful cost (~0.5–1 ms per round-trip). `ffiMode=raw` skips serialisation entirely — flat C structs cross via shared-memory pointers. `cbor` remains the universal/portable choice and works in-process too.

## What changed

- **2-axis selection**: `ffiMode` (`raw` | `cbor`) + `ffiLang` (`cpp` | `rust`) replace `targetLang`. Dispatch on `(ffiMode, ffiLang)` in `genBindings`.
- **Content-addressed symbols** (raw only): exportc symbol is `<name>_v<hash>` where `<hash>` is a 7-char FNV-1a prefix of the canonical signature (param names/types, transitively through nested `{.ffi.}` types, plus return type). Any structural change → new symbol → stale consumers fail to link with `undefined symbol` instead of silently corrupting memory. Header emits `#define <clean> <hashed>` so consumer source stays clean. Dtor is structurally invariant and stays unsuffixed.
- **Param-order alignment**: both modes share the leading prefix `cint; ctx, callback, userData, …`; only the trailing payload params differ (CBOR: `reqCbor, reqCborLen`; raw: `req: ptr <ReqType>_CWire`).
- **Nimble tasks** renamed to `genbindings_<lang>_<mode>` (`genbindings_cpp_cbor`, `genbindings_cpp_rawc`, `genbindings_rust_cbor`). Legacy `genbindings_cpp` / `genbindings_rust` / `genbindings_c` aliases removed.
- **README** rewritten with the matrix, per-value explanations, and a decision rule (same-process + per-call cost matters → `raw`; portable / ABI-tolerant → `cbor`, even in-process).

## Test plan

- [x] **Unit** 24/24 under `--mm:orc` + `--mm:refc` (12 files × 2 GC modes), including 11 new ABI-hash tests + 4 new C-header emission tests.
- [x] **C e2e** 6/6 via `nimble test_c_e2e` (smoke, cross-library, 4 stress scenarios).
- [x] **C++ e2e** 9/9 via `nimble test_cpp_e2e` — CBOR path unaffected.
- [x] Matrix sweep: every cell builds; `(raw, rust)` rejects with a clean "not yet implemented" error.

## Backward compatibility

Breaking surface change:

- `targetLang` strdefine **removed** — use `-d:ffiMode=… -d:ffiLang=…`.
- `genbindings_cpp` / `genbindings_rust` / `genbindings_c` aliases **removed** — use the explicit `genbindings_<lang>_<mode>` task names.
- Raw-mode exported C symbols now carry a `_v<hash>` suffix; consumer source picks this up via the generated `#define`, but pre-built binaries built against the old surface need a rebuild.

The CBOR path is otherwise unchanged.


ffi/codegen/templates/c/CMakeLists.txt.tpl
```diff
@@ -0,0 +1,68 @@
...
+# ── Locate the repository root (contains ffi.nimble) ─────────────────────────
+set(_search_dir "${CMAKE_CURRENT_SOURCE_DIR}")
+set(REPO_ROOT "")
+foreach(_i RANGE 10)
+    if(EXISTS "${_search_dir}/ffi.nimble")
+        set(REPO_ROOT "${_search_dir}")
+        break()
+    endif()
+    get_filename_component(_search_dir "${_search_dir}" DIRECTORY)
+endforeach()
```
gmelodie, line 16: Why do this 10 times? This needs to be more robust/deterministic

ffi/codegen/templates/c/CMakeLists.txt.tpl
```diff
@@ -0,0 +1,68 @@
...
+if("${REPO_ROOT}" STREQUAL "")
+    message(FATAL_ERROR "Cannot find repo root (no ffi.nimble in any ancestor)")
+endif()
```
gmelodie, line 19: Should you check `_search_dir` directly in the begining instead?

ffi/codegen/templates/c/CMakeLists.txt.tpl
```diff
@@ -0,0 +1,68 @@
...
+                -d:ffiMode=raw
+                -d:ffiLang=cpp
+                --app:lib
+                --noMain
+                "--nimMainPrefix:lib{{LIB}}"
+                "--nimcache:${CMAKE_CURRENT_BINARY_DIR}/nimcache_{{LIB}}_c"
+                "-o:${NIM_LIB_FILE}"
+                "${NIM_SRC}"
+    WORKING_DIRECTORY "${REPO_ROOT}"
+    DEPENDS "${NIM_SRC}"
```
gmelodie, line 49: Do you not also need `DEPENDS ${CMAKE_CURRENT_BINARY_DIR}`? Is that always set by cmake?

ffi/codegen/templates/c/header_prelude.h.tpl
```diff
@@ -0,0 +1,31 @@
...
+ * FFI callback type.
+ *
+ * Used as the success/failure notification for every async call. Identical
+ * to the Nim `FFICallBack` signature byte-for-byte; reused unchanged by the
+ * C++ sugar layer.
+ *
+ *   ret      : 0 = OK, 1 = error, 2 = missing callback
+ *   msg      : on RET_OK, points to a `const <Proc>Resp*` (caller must cast).
+ *              On RET_ERR, points to a UTF-8 error message of `len` bytes
+ *              (not necessarily NUL-terminated).
```
gmelodie, line 21: Why not?

ffi/codegen/c.nim
```diff
@@ -0,0 +1,284 @@
...
+import std/[os, strutils]
+import ./meta, ./string_helpers
+
+## Static template blocks live as real C / CMake files under templates/c/
+## and are slurped into the binary at compile time.
+const
+  HeaderPreludeTpl = staticRead("templates/c/header_prelude.h.tpl")
+  CMakeListsTpl = staticRead("templates/c/CMakeLists.txt.tpl")
+
+proc genericInnerType(typeName, prefix: string): string =
```
gmelodie, line 20: nit 
```suggestion
proc extractInnerType(typeName, prefix: string): string =
```

ffi/codegen/c.nim
```diff
@@ -0,0 +1,284 @@
...
+  ## If `typeName` looks like `Prefix[Inner]`, return `Inner`. Otherwise "".
+  if typeName.startsWith(prefix) and typeName.endsWith("]"):
+    let start = prefix.len
+    let lastIndex = typeName.len - 2
+    return typeName[start .. lastIndex]
+  return ""
+
+proc nimTypeToC*(typeName: string): string =
```
gmelodie, line 28: nit 
```suggestion
proc nimTypeToCType*(typeName: string): string =
```

ffi/codegen/c.nim
```diff
@@ -0,0 +1,284 @@
...
+  ## Map a Nim type name (as captured by FFIFieldMeta.typeName) to its C form.
+  ## Mirrors the cpp.nim mapper but emits plain C.
+  let trimmed = typeName.strip()
+  if trimmed.startsWith("ptr "):
+    return "void*"
```
gmelodie, line 33: Can we not sometimes infer the type? e.g.
```nim
var x: int = 42
var p: ptr int = addr x
```
The above would be `int *`

So maybe the approach would be: if we can deduct, do it, otherwise use `void *` as a fallback

ffi/codegen/c.nim
gmelodie: Also we could instead return an `Opt` so that you could do 
```nim
extractInnerType(yourType).withValue(innerType):
  # do something
 ```

File: ffi/codegen/c.nim [hunk 8]
```diff
@@ -0,0 +1,284 @@
...
+  lines.add("#endif")
+  lines.add("")
+
+  return lines.join("\n")
+
+proc generateCCMakeLists*(libName: string, nimSrcRelPath: string): string =
+  let src = nimSrcRelPath.replace("\\", "/")
+  return CMakeListsTpl.multiReplace(("{{LIB}}", libName), ("{{SRC}}", src))
+
+proc cHeaderBanner(nimSrcRelPath: string): string =
```
gmelodie, line 247:
```suggestion
proc generateCHeaderBanner(nimSrcRelPath: string): string =
```

ffi/codegen/c.nim
```diff
@@ -0,0 +1,284 @@
...
+  ## that it is auto-produced and where to look to regenerate.
+  return
+    "/* ============================================================\n" &
+    " * AUTO-GENERATED by nim-ffi v" & NimFFIVersion &
+    "  (ffiMode=raw, ffiLang=cpp). DO NOT EDIT.\n" & " *\n" & " *   Source:     " &
+    nimSrcRelPath & "\n" & " *   Generator:  ffi/codegen/c.nim\n" &
+    " *   Regenerate: nimble genbindings_cpp_rawc\n" &
+    " * ============================================================ */\n\n"
+
+proc cMakeBanner(nimSrcRelPath: string): string =
```
gmelodie, line 259:
```suggestion
proc generateCMakeBanner(nimSrcRelPath: string): string =
```

ffi/codegen/meta.nim
```diff
@@ -34,12 +49,191 @@ var ffiProcRegistry* {.compileTime.}: seq[FFIProcMeta]
...
 # Output directory for generated bindings; set with -d:ffiOutputDir=path/to/dir
 const ffiOutputDir* {.strdefine.} = ""
 
 # Nim source path (relative to outputDir) embedded in generated build files;
 # set with -d:ffiSrcPath=../relative/path.nim
 const ffiSrcPath* {.strdefine.} = ""
+
+proc effectiveMode*(): string {.compileTime.} =
+  ## Returns the active ABI mode (always lowercased). Defaults to `cbor`.
+  return ffiMode.toLowerAscii()
```
gmelodie, line 98: !!!!!!!!!!!!!!!!!
```suggestion
  return ffiMode.toLower()
```

ffi/codegen/meta.nim
```diff
@@ -34,12 +49,191 @@ var ffiProcRegistry* {.compileTime.}: seq[FFIProcMeta]
...
+# source of truth.
+# ---------------------------------------------------------------------------
+
+const NimblePkgVersion* {.strdefine.} = ""
+
+proc parseNimbleVersion(): string {.compileTime.} =
+  const nimbleContent = staticRead("../../ffi.nimble")
+  for line in nimbleContent.splitLines():
+    let s = line.strip()
+    if s.startsWith("version"):
```
gmelodie, line 122: The way this is written will match things like `versionTest = "1"`
```suggestion
    if s.startsWith("version ") or s.startsWith("version="):
```

ffi/codegen/meta.nim
```diff
@@ -34,12 +49,191 @@ var ffiProcRegistry* {.compileTime.}: seq[FFIProcMeta]
...
+const AbiHashPrefixLen* = 7
+
+proc canonicalTypeName*(typeName: string, depth: int = 0): string {.compileTime.} =
+  ## Canonical, recursively-expanded representation of a type for ABI
+  ## hashing. Primitives map to themselves; `seq[T]` / `Option[T]` /
+  ## `Maybe[T]` expand to `seq[<canonical T>]` etc.; nested {.ffi.} object
+  ## types expand to `TypeName{field1:type1;field2:type2;...}` so the
+  ## transitive shape is part of the hash. `Maybe` aliases to `Option` so
+  ## the two read identically.
+  if depth > 16:
```
gmelodie, line 164: Why was 16 chosen?

ffi/codegen/meta.nim
```diff
@@ -34,12 +49,191 @@ var ffiProcRegistry* {.compileTime.}: seq[FFIProcMeta]
...
+    return "<recursive>"
+  let t = typeName.strip()
+  case t
+  of "string", "cstring", "int", "int8", "int16", "int32", "int64", "uint", "uint8",
+      "uint16", "uint32", "uint64", "bool", "float", "float32", "float64", "pointer",
+      "void":
```
gmelodie, line 170: We could make this a `const` list of strings and do 
```nim
if t in CannonicalTypes: 
  return t
```

ffi/codegen/meta.nim
```diff
@@ -34,12 +49,191 @@ var ffiProcRegistry* {.compileTime.}: seq[FFIProcMeta]
...
+    return t
+  else:
+    discard
+  if t.startsWith("seq[") and t.endsWith("]"):
```
gmelodie, line 174: Create an `isSeq` proc for this (similarly for `Opt` etc below), reuse these functions where possible, not only here

ffi/codegen/meta.nim
```diff
@@ -34,12 +49,191 @@ var ffiProcRegistry* {.compileTime.}: seq[FFIProcMeta]
...
+  ## can't execute). Good enough as a fingerprint for distinguishing
+  ## struct shapes — not a cryptographic primitive, but we're not
+  ## defending against an adversary, just detecting drift.
+  const Offset = 0xcbf29ce484222325'u64
+  const Prime = 0x100000001b3'u64
+  var h = Offset
+  for c in s:
+    h = h xor uint64(byte(c))
+    h = h * Prime
+  return h
```
gmelodie, line 209: wth does this do?

ffi/internal/c_macro_helpers.nim
```diff
@@ -0,0 +1,578 @@
...
+##   - `cwireUnpack(src: T_CWire): T`          — copies back into Nim heap
+##   - `cwireFree(dst: var T_CWire)`           — releases shared cstrings
+##
+## Wire structs live in `allocShared` memory so they can travel from the
+## C caller's thread through the FFI channel without crossing GC boundaries.
+
+import std/[macros, strutils, tables]
+import ../codegen/meta
+
+proc isPureCMode*(): bool {.compileTime.} =
```
gmelodie, line 24:
```suggestion
proc isRawMode*(): bool {.compileTime.} =
```

ffi/internal/c_macro_helpers.nim
```diff
@@ -0,0 +1,578 @@
...
+# ---------------------------------------------------------------------------
+# Compile-time registry of known ffi-type names → used to recognise nested
+# user types so the wire form can use their `_CWire` companion.
+# ---------------------------------------------------------------------------
+
+proc knownFFITypeNames*(): seq[string] {.compileTime.} =
+  var names: seq[string] = @[]
+  for t in ffiTypeRegistry:
+    names.add(t.name)
+  return names
```
gmelodie, line 43: This is not used anywhere

ffi/internal/c_macro_helpers.nim
```diff
@@ -0,0 +1,578 @@
...
+  ## Linear scan over the emission tracker. Written as an indexed loop to
+  ## sidestep a Nim 2.2 compile-time VM quirk where repeated `for x in
+  ## seq` reads of a freshly-mutated `{.compileTime.}` seq can return
+  ## stale results within a single macro expansion.
+  var found = false
+  for i in 0 ..< emittedCWireTypes.len:
+    if emittedCWireTypes[i] == typeName:
+      found = true
+      break
+  return found
```
gmelodie, line 72: Wait so you can't have 2 or more `for x in seq` nested loops in macros? If this is not a known bug we should open an issue on the compiler

ffi/internal/c_macro_helpers.nim
```diff
@@ -0,0 +1,578 @@
...
+    emittedCWireTypes.add(typeName)
+
+# ---------------------------------------------------------------------------
+# Type-level helpers
+# ---------------------------------------------------------------------------
+
+proc cwireTypeName*(userTypeName: string): string =
+  ## Companion-type naming convention. Kept stable so generated tests can
+  ## reach in by name.
+  return userTypeName & "_CWire"
```
gmelodie, line 85: Please explain better

ffi/internal/c_macro_helpers.nim
```diff
@@ -0,0 +1,578 @@
...
+
+proc isStringType(t: NimNode): bool =
```
gmelodie, line 87: nit

same for other occurrences
```suggestion
proc isString(t: NimNode): bool =
```

ffi/internal/c_macro_helpers.nim
```diff
@@ -0,0 +1,578 @@
...
+  t.kind == nnkIdent and ($t == "string" or $t == "cstring")
+
+proc isSeqType(t: NimNode): bool =
+  t.kind == nnkBracketExpr and t.len >= 2 and t[0].kind == nnkIdent and $t[0] == "seq"
+
+proc isOptionType(t: NimNode): bool =
+  t.kind == nnkBracketExpr and t.len >= 2 and t[0].kind == nnkIdent and
+    ($t[0] == "Option" or $t[0] == "Maybe")
+
+proc isNestedFFIType(t: NimNode): bool =
```
gmelodie, line 97: Split in two: `isNested` and `isFFIType`

ffi/internal/c_macro_helpers.nim
```diff
@@ -0,0 +1,578 @@
...
+  t.kind == nnkIdent and isKnownFFIType($t)
+
+proc wireFieldType(userType: NimNode): NimNode =
+  ## Map a user field type AST → its wire-form AST.
+  ## Note: `seq` is not handled here because it must split into two physical
+  ## fields; the caller uses `wireFieldsFor` for that case.
+  if isStringType(userType):
```
gmelodie, line 104: The types here could be an `enum` and this function could be only a big `case...of`

ffi/internal/c_macro_helpers.nim
```diff
@@ -0,0 +1,578 @@
...
+      if isNestedFFIType(elem):
+        ident(cwireTypeName($elem))
+      else:
+        wireFieldType(elem)
+    let itemsField = newIdentDefs(
+      ident(fieldName & "_items"),
+      nnkPtrTy.newTree(nnkBracketExpr.newTree(ident("UncheckedArray"), elemWire)),
+      newEmptyNode(),
+    )
+    let lenField = newIdentDefs(ident(fieldName & "_len"), ident("int"), newEmptyNode())
```
gmelodie, line 135: This whole block is inside `if isSeqType`, it'd be better to invert the logic to avoid identation. Same for other occurrences.

ffi/internal/c_macro_helpers.nim
```diff
@@ -0,0 +1,578 @@
...
+
+proc buildCWireTypeDef*(
+    userTypeName: string, fieldNames: seq[string], fieldTypes: seq[NimNode]
+): NimNode =
+  ## Build the bare `nnkTypeDef` (no enclosing TypeSection) for the wire
+  ## companion of `userTypeName`. Callers splice the result into whichever
+  ## TypeSection makes sense at the call site.
+  let wireName = ident(cwireTypeName(userTypeName))
+  var fields: seq[NimNode] = @[]
+  for i in 0 ..< fieldNames.len:
```
gmelodie, line 147: not sure if you can do this in nim
```suggestion
  for fdName, fdType in zip(fieldNames, fieldTypes):
```

ffi/internal/c_macro_helpers.nim
```diff
@@ -0,0 +1,578 @@
...
+    let idx = genSym(nskForVar, "i")
+    let elemUnpack =
+      emitElemUnpack(elemVar, nnkBracketExpr.newTree(itemsField, idx), elemType)
+    stmts.add quote do:
+      `dstAccess` = @[]
+      for `idx` in 0 ..< `lenField`:
+        var `elemVar`: `elemType`
+        `elemUnpack`
+        `dstAccess`.add(`elemVar`)
+    return stmts
```
gmelodie, line 301: These long parsings inside `if`statements should be extracted to functions to help with readability

ffi/internal/c_macro_helpers.nim
```diff
@@ -0,0 +1,578 @@
...
+  let userName = ident(userTypeName)
+  let wireName = ident(cwireTypeName(userTypeName))
+
+  # --- cwirePack(dst: var T_CWire, src: T) ---
+  let packDst = ident("dst")
+  let packSrc = ident("src")
+  var packBody = newStmtList()
+  for i in 0 ..< fieldNames.len:
+    let fIdent = ident(fieldNames[i])
+    for s in emitPackStmt(packDst, packSrc, fIdent, fieldTypes[i]):
```
gmelodie, line 385: Same as above: `zip` `fieldNames` and `fieldTypes` together

ffi/internal/c_macro_helpers.nim
```diff
@@ -0,0 +1,578 @@
...
+  for i in 0 ..< fieldNames.len:
+    let fIdent = ident(fieldNames[i])
+    for s in emitUnpackStmt(unpRes, unpSrc, fIdent, fieldTypes[i]):
+      unpBody.add(s)
+  unpBody.add quote do:
+    return `unpRes`
+  let unpProc = newProc(
+    name = postfix(ident("cwireUnpack"), "*"),
+    params = @[userName, newIdentDefs(unpSrc, wireName)],
+    body = unpBody,
```
gmelodie, line 415: The fact that there's this `---` comment above tells me really clearly that this should be a separate proc, same for other occurrences below.

ffi/internal/c_macro_helpers.nim
```diff
@@ -0,0 +1,578 @@
...
+  if not isPureCMode():
+    return (nil, @[])
+  result.wireType = buildCWireTypeDef(reqTypeName, paramNames, paramTypes)
+  result.procs = buildCWireProcs(reqTypeName, paramNames, paramTypes)
+
+# ---------------------------------------------------------------------------
+# Helper to collect nested type dependencies (transitively).
+# ---------------------------------------------------------------------------
+
+proc collectNestedFFITypes*(
```
gmelodie, line 517: We could benefit from stronger typing here:
```nim
type FFIType = NimNode
```
then you could have
```nim
type Nested = FFIType
```
this way you could have the check for `FFIType` when creating the type, and only test if the type is nested when casting it to `Nested`. Another possible design is to have `FFIType` be an `enum` and `Nested` be a variant of that enum.

ffi/internal/c_macro_helpers.nim
```diff
@@ -0,0 +1,578 @@
...
+    if isNestedFFIType(t):
+      let n = $t
+      if n notin deps:
+        deps.add(n)
+    elif isSeqType(t) or isOptionType(t):
+      let inner = t[1]
+      if isNestedFFIType(inner):
+        let n = $inner
+        if n notin deps:
+          deps.add(n)
```
gmelodie, line 534: Follow up to the comment above, this could be instead turned into

```suggestion
    case t:
    of FFIType.Nested:
      let n = $t
      if n notin deps:
        deps.add(n)
    of FFIType.Seq:
    of FFIType.Opt:
    ...
```

ffi/internal/c_macro_helpers.nim
```diff
@@ -0,0 +1,578 @@
...
+
+proc fieldInfoForType(
+    typeName: string
+): tuple[names: seq[string], types: seq[NimNode]] {.compileTime.} =
+  ## Look up an ffi type's fields from the compile-time registry and parse
+  ## each field's recorded type back into a NimNode AST.
```
gmelodie, line 540: This could also be part of the `FFIType` field (if it were an object instead of an enum, or maybe create another type like `FFITypeInfo` that has `name` and `type`. Or even
```nim
type
  FFITypeKind = enum
    Nested, Opt, Seq

  FFIType = object
    case kind: FFITypeKind
    of Nested:
      name: string
      `type`: NimNode
    of Opt:
      `type`: NimNode
    of Seq:
      `type`: NimNode
```
The pattern above lets different types of `FFIType`s have different fields if need be

ffi/internal/c_macro_helpers.nim
```diff
@@ -0,0 +1,578 @@
...
+  for tm in ffiTypeRegistry:
+    if tm.name == typeName:
+      for f in tm.fields:
+        result.names.add(f.name)
```
gmelodie, line 544: No `result`s!!

ffi/ffi_thread_request.nim
```diff
@@ -1,6 +1,25 @@
...
+##     The response, similarly, is a `Resp_CWire` ptr cleaned up after the
+##     callback fires.
+##
+## The C-wire mode bolts onto the existing carrier via two cleanup hooks:
+##   - `cleanupReqProc`  — invoked by `deleteRequest` so generated code can
+##     free cstrings/arrays owned by the wire Req struct.
+##   - `cleanupRespData` / `cleanupRespProc` — set by the FFI-thread handler
+##     before returning the response bytes; `handleRes` invokes the proc
+##     **after** firing the user callback so any shared-memory strings
+##     pointed to by the Resp survive at least until the callback returns.
```
gmelodie, line 22: There's way too many and too verbose comments, which tells me the code is more cryptic than it should be, or that the explanation should be on proper docs somewhere else, or maybe even on the examples.

tests/test_c_codegen.nim
```diff
@@ -0,0 +1,325 @@
+## Unit tests for the pure-C binding generator at `ffi/codegen/c.nim`.
+##
```
gmelodie, line 2:
```suggestion
```

tests/test_c_codegen.nim
```diff
@@ -0,0 +1,325 @@
...
+## The generator is a pure-input → pure-output function: it takes the
+## compile-time `FFIProcMeta` / `FFITypeMeta` registries plus a library
+## name and returns a header string. That makes it straightforward to
+## drive without touching the runtime — these tests construct synthetic
+## metadata and assert structural properties on the emitted text.
```
gmelodie, line 7: not test-related, it explains the code, shouldn't be here
```suggestion
```

tests/test_c_codegen.nim
```diff
@@ -0,0 +1,325 @@
...
+  "string",
+)
+const renamedParamHash =
+  hashFor("foo", @[FFIParamMeta(name: "x", typeName: "int")], "string")
+const retypedParamHash =
+  hashFor("foo", @[FFIParamMeta(name: "a", typeName: "int32")], "string")
+const renamedProcHash =
+  hashFor("bar", @[FFIParamMeta(name: "a", typeName: "int")], "string")
+const changedRetHash =
+  hashFor("foo", @[FFIParamMeta(name: "a", typeName: "int")], "int")
```
gmelodie, line 238: move these inside each pertinent test

tests/test_c_macro_helpers.nim
```diff
@@ -0,0 +1,171 @@
+## Unit tests for `ffi/internal/c_macro_helpers.nim`. The module is mostly
+## compile-time AST builders, so the tests assert structural invariants on
+## the produced ASTs via `treeRepr` substring matches. The tracking
+## registry (`emittedCWireTypes`) is also covered because it underpins
+## ordering correctness in the cwire-emission pipeline.
+
```
gmelodie, line 6:
```suggestion
```

gmelodie (commented):
1. in `ffi/codegen/c.nim` I think it'd be good to have small functions for every type (e.g. `seqToC`, `constToC`) so that `nimTypeToC` could simply be a huge `switch...case` 
2. in `ffi/codegen` generation functions should all be called `generateC...` or `generateCPP`, etc. to maintain standard 
3. for codegen functions that are too similar like generating banners, etc, maybe it'd be good to create a `utils.nim` file with all of those, or have separate dirs for each language so we could break down the files and make them more manageable.
4. This PR could easily be broken down into smaller ones, unit testing conversion and macros piece by piece. Once all the pieces were in place after the smaller PRs had been merged, then a final PR that uses all that pre-merged code could be introduced. That final PR would be easier to read and reason about. Maybe we could add a hard limit of around 500 LOC for a PR. If a PR exceeds that, it should probably be broken down except it really unique cases.
5. There's many comments in code, and very verbose ones. That usually is a hint that the code does not explain itself very well and should be refactored and simplified.

---
Question: the `FNV-1a 64-bit` hashing-related code, is that used for name mangling? Please explain 
  
---
Obs: maybe we could introduce a couple instructions for AI using a CLAUDE.md file in the repo for things such as "always prefer `toLower` instead of `toLowerAscii`" and "never use the `result` keyword"

NagyZoltanPeter:
@Ivansete-status @gmelodie 
I have a generic question here: So is it the plan to have FFI surface all wired into a struct / call?

Just a warning note: in nim-brokers I'm just about to retire native functional C-ABI option.
I have good reason for that. First is that compared to CBor it is not possible/easy to have a complete mapping of Nim structs/constructs to C. It would need too complex memory management that is error prone.
Also at least Nim 2.2.4 allocator has serious race issue tied to small object allocator optimization. 
I created benchmark comparison and turned out CBOR buffering is performing at least that good but for complex constructs that involves object hiearachy or seq[T]... and such things CBOR just simple outperforms native format.
It is because you need to manage (or at least it was true for RequestBroker's) recursive allocations on thread boundary.
It can be problematic when thread teardown happen while having in-flight transports. (use after free, allocator race conditions, ... )
In short it was simple not worth doing so.
I think C-ABI can be supported derived from CBOR buffering (of course the compilation will not that simple due need of tinycbor or similar dependency).

Also worth to consider, CBOR tunneling will always be ABI stable even if surface changes, because of marshalling mesh away the change on the Nim surface. Old client can still call with new build, if the existed API did not changed.
In other word it allows API versioning too.

ffi/codegen/meta.nim
Ivansete-status: That implements a hash algorithm (surely is done in another lib.)
Its purpose is to uniquely identify any exposed function, i.e., mangling, so that in case anything changes in a func signature (types, types' order, etc.) then we have a different symbol that will cause a big compilation error instead of silently compile and produce an UB.

Ivansete-status:
Thanks so much for the feedback @NagyZoltanPeter @gmelodie !
I will pause the work on that until we bump nim-ffi and use its cbor implementation.
I'll revisit that work in the future.
For now, a simple benchmark to compare both performances: https://github.com/logos-messaging/nim-ffi/tree/benchmark-cbor-c/tests/bench#reading-the-numbers

Ivansete-status:
Closing this PR as it is too complex. This will be completed in the following issue: https://github.com/logos-messaging/nim-ffi/issues/78

PR #34, opened 2026-05-19
chore(ci): fsanitize tests

### LINKED ISSUE #32: Add sanitizer tests
Add dedicated CI jobs (separate from the default cpp-e2e job because of slowdown + TSan/ASan incompatibility):

- cpp-e2e-asan-ubsan — CXXFLAGS=-fsanitize=address,undefined -fno-sanitize-recover=all, Linux, GCC + Clang.
- cpp-e2e-lsan — LeakSanitizer (Linux). Suppress only known Nim-runtime leaks via a checked-in lsan.supp.
- cpp-e2e-tsan — ThreadSanitizer over ThreadedHammer and the async cases. Fail the job on any sanitizer report.

gmelodie:
Sanitizer CI: run all tests under ASan+UBSan+LSan and TSan

Adds two umbrella CI jobs that run every test (unit + C++ e2e) under sanitizers, separate from the fast-feedback path:

- **Tests · ASan+UBSan+LSan** — ASan with `detect_leaks=1` (LSan rides ASan's runtime) + UBSan
- **Tests · TSan** — ThreadSanitizer

Matrix per job: `{ubuntu-22.04, ubuntu-24.04-arm} × {orc, refc}` = 4 cells. Single compiler (runner-default GCC), pinned Nim 2.2.4. `fail-fast: false`.

### Drive-by fix: `allocShared`/`createShared` → `c_malloc`/`c_free` in `ffi_thread_request.nim`

The sanitizer run surfaced a real use-after-free in the thread-request allocator under `--mm:orc`. Nim's `allocShared`/`createShared` are backed by a **per-thread `MemRegion` stored in TLS**. The thread that calls `allocShared` owns the chunk, and the chunk's `owner` pointer references that thread's TLS region. When `deallocShared`/`freeShared` runs later, it follows `owner` back into TLS.

The producer in our flow is typically a transient `std::async` worker on the foreign side that exits as soon as it's queued the request. By the time the FFI thread runs `deleteRequest`, the producer's TLS has been reclaimed — `owner` dangles, and the deallocator segfaults (the bug fixed in e27af2a).

Switching the envelope, the `reqId` copy, and the payload buffer to libc `malloc`/`free` removes the TLS coupling entirely: those allocators are process-global, so any thread can free what any other thread allocated. `refc` doesn't hit this because its shared allocator isn't TLS-backed — but the asymmetry was the bug, not a property worth preserving.

Fixes #32.


ffi/ffi_thread_request.nim
```diff
@@ -21,10 +30,10 @@ type FFIThreadRequest* = object
...
     callback: FFICallBack, userData: pointer, reqId: cstring
 ): ptr FFIThreadRequest =
-  ## Allocates the request envelope in shared memory and populates the
-  ## routing fields. Payload setup is delegated to one of the payload helpers
-  ## below depending on whether the bytes need to be copied or adopted.
-  var ret = createShared(FFIThreadRequest)
+  ## Allocates the request envelope via `c_malloc` and populates the routing
+  ## fields. Payload setup is delegated to one of the payload helpers below
+  ## depending on whether the bytes need to be copied or adopted.
+  var ret = cast[ptr FFIThreadRequest](c_malloc(csize_t(sizeof(FFIThreadRequest))))
```
Ivansete-status, line 36: What was the problem with allowShared and createShared? It seems problem can be caused when using `orc`. Kindly elaborate that change in the PR's description.

versions.env
```diff
@@ -0,0 +1,2 @@
+NIM_VERSIONS=["2.2.4","stable"]
```
Ivansete-status, line 1: That's fine but I wonder if we should move this to a more visible place such as root folder, next to ffi.nimble.

.github/workflows/tests-sanitized.yml
```diff
@@ -0,0 +1,92 @@
...
+          key: ${{ runner.os }}-${{ runner.arch }}-nimbledeps-${{ matrix.nim-version }}-${{ hashFiles('*.nimble') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ runner.arch }}-nimbledeps-${{ matrix.nim-version }}-
+            ${{ runner.os }}-${{ runner.arch }}-nimbledeps-
+
+      - name: Install nimble deps
+        if: steps.cache-nimbledeps.outputs.cache-hit != 'true'
+        run: nimble setup --localdeps -y
+
+      - name: Cache CMake FetchContent (GoogleTest + nlohmann_json)
```
Ivansete-status: I might be missing something but, it seems `nlohmann` is not used.

ffi.nimble
```diff
@@ -16,6 +16,44 @@ requires "cbor_serialization"
...
+  "test_ffi_context",
+  "test_gc_compat",
+  "test_serial",
+  "test_ctx_validation",
+  "test_nim_native_api",
+  "test_meta",
+  "test_string_helpers",
+  "test_wire_compat",
+  "test_cddl_codegen",
+]
```
Ivansete-status:
```suggestion
proc discoverUnitTests(): seq[string] =
  var names: seq[string] = @[]
  for path in listFiles(thisDir() / "tests/unit"):
    let name = path.extractFilename.changeFileExt("")
    if name.startsWith("test_"):
      names.add(name)
  names.sort()
  return names

let unitTests = discoverUnitTests()
```
With that, we can also have the following a few lines below:
```nim
task test, "Run all tests under --mm:orc and --mm:refc":
  for flags in [nimFlagsOrc, nimFlagsRefc]:
    for t in unitTests:
      exec "nim c -r " & flags & " tests/unit/" & t & ".nim"
```

Ivansete-status (approved):
LGTM! Thanks for it! 🙌

ffi/ffi_thread_request.nim
gmelodie: Added to pr description

PR #35, opened 2026-05-19
chore: bump nim to 2.2.6 and ci to 2.2.10

gmelodie:
Not sure if this is sth we want, but it wasn't much effort to create this.


Ivansete-status:
Thanks for that @gmelodie ! We stick to 2.2.4 as the minimum version because is the one we use in `logos-delivery`. At some point, not urgent, we might need to bump Nim compiler there.

PR #36, opened 2026-05-19
avoid move ctor and assing operator in cpp generated code

Ivansete-status:
This tackles this comment from @NagyZoltanPeter ( thanks 🙌 ! )
https://github.com/logos-messaging/nim-ffi/commit/4dffdaa86c22384a93e95cf27196d8f9d3f2d258#r185483589

The main idea is to avoid auto-generate move ctor and assignment operator for Cpp. Also, wrapping the context instance in an `unique_ptr<>` so that we make it as movable as in Rust.
Go, Python, Java, C, Swift don't have such "move" concept so major targets might get similar behaviour.

### DIFF
--- a/examples/timer/cpp_bindings/main.cpp
+++ b/examples/timer/cpp_bindings/main.cpp
@@ -7,9 +7,9 @@ int main() {
         auto ctx = MyTimerCtx::create(TimerConfig{"cpp-demo"});
         std::cout << "[1] Context created\n";
 
-        auto versionFuture = ctx.versionAsync();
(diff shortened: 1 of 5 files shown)


fcecin (approved):
LGTM

NagyZoltanPeter (approved):
Yepp, wise move

gmelodie (approved):
Nice find!

Ivansete-status:
All kudos to @NagyZoltanPeter 🙌 ! Thanks !

PR #38, opened 2026-05-21
chore(ci): extend cpp-e2e to OS matrix

### LINKED ISSUE #33: Extend e2e test platform matrix
Extend cpp-e2e from ubuntu-22.04 only to a strategy matrix {ubuntu-22.04, macos-14, windows-latest}. Windows likely needs MSVC-specific tweaks in the generated CMake (/EHsc, dllexport); track follow-ups inline if the port surfaces codegen bugs.

gmelodie:
Add macos-15 and windows-latest to the cpp-e2e job alongside the existing ubuntu-22.04 runner. Steps now use `shell: bash` with the same RUNNER_OS=Windows PATH shims as test.yml. fail-fast stays false so a Windows failure doesn't mask Linux/macOS results.

closes #33

### DIFF
--- a/ffi.nimble
+++ b/ffi.nimble
@@ -33,6 +33,18 @@ proc discoverUnitTests(): seq[string] =
 
 let unitTests = discoverUnitTests()
 
+proc runOrQuit(cmd: string) =
+  # Workaround for newer nimble (shipping with Nim 2.2.10+) printing the
+  # OSError from a failed `exec` but exiting 0, which causes CI to report
+  # green on actual build/test failures. Echo the command first so the log
+  # makes clear which step failed.
+  try:
+    exec cmd
+  except OSError as e:
+    echo "command failed: ", cmd
+    echo "error: ", e.msg
+    quit(QuitFailure)
+
 proc sanFlags(san: string): string =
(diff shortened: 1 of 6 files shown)


ffi.nimble
```diff
@@ -80,10 +80,16 @@ task test_serial, "Run CBOR codec unit tests":
...
+  # The try/except + quit is a workaround for newer nimble (shipping with
+  # Nim 2.2.10+) printing the OSError from a failed `exec` but exiting 0,
+  # which causes CI to report green on actual build/test failures.
+  try:
+    exec "nimble genbindings_cpp"
+    exec "cmake -S tests/e2e/cpp -B tests/e2e/cpp/build"
+    exec "cmake --build tests/e2e/cpp/build"
+    exec "ctest --test-dir tests/e2e/cpp/build --output-on-failure"
+  except OSError:
+    quit(QuitFailure)
```
Ivansete-status: I wonder if we could bring better detail about where the exception was raised, i.e., which exec. Applies elsewhere.

Ivansete-status (approved):
LGTM! Thanks for it! 🙌

PR #39, opened 2026-05-21
Adjust events to cbor

Ivansete-status:


fcecin (approved):

ffi/ffi_context.nim
```diff
@@ -83,14 +86,43 @@ template dispatchFFIEvent*(eventName: string, body: untyped) =
...
+  ##
+  ## NB: the template parameter is intentionally named `eventPayload`
+  ## rather than `payload` — Nim's template substitution would otherwise
+  ## also replace the `payload:` field name inside `EventEnvelope`.
+  dispatchFFIEvent(eventName):
+    cborEncode(
+      EventEnvelope[typeof(eventPayload)](
+        eventType: eventName, payload: eventPayload
+      )
+    )
```
NagyZoltanPeter: I looked this up deeper and it runs on processing thread (which was the case before, and my guess it will be addressed in the future), but it passes a nim gc owned memory over the C-ABI boundary, which seems safe - if we take that our generated code will only use it - but still would be wiser to use `cborEncodeShared` and handle freeing it. Assumption that it will not be used from another thread and the reference may outlive the processing thread sounds weak.

ffi/codegen/cpp.nim
```diff
@@ -135,8 +135,91 @@ proc cppBracedInit(structName: string, fieldNames: seq[string]): string =
...
+  lines.add("        std::function<void(const std::string&)> on_error;")
+  for ev in events:
+    lines.add(
+      "        std::function<void(const $1&)> $2;" %
+        [ev.payloadTypeName, ev.nimProcName]
+    )
+  lines.add("    };")
+  lines.add("")
+  lines.add("    void setEventHandlers(Events handlers) {")
+  lines.add("        events_ = std::make_unique<Events>(std::move(handlers));")
```
NagyZoltanPeter, line 164: What happens with the events eventually already set?
So to say the lib context at this point still thinks the old handlers are the valid ones and may call them midflight.

ffi/codegen/cpp.nim
```diff
@@ -135,8 +135,91 @@ proc cppBracedInit(structName: string, fieldNames: seq[string]): string =
...
+  lines.add(
+    "        $1_set_event_callback(ptr_, &$2::eventTrampoline, events_.get());" %
```
NagyZoltanPeter, line 166: This might be not the appropriate place to mention but this function call is unguarded, it changes libcontext from another thread.
`ctx[].callbackState` must be protected from concurrent writes and dirty reads.

ffi/ffi_context.nim
```diff
@@ -66,14 +68,15 @@ template callEventCallback*(ctx: ptr FFIContext, eventName: string, body: untype
...
         ctx[].callbackState.userData,
       )
 
 template dispatchFFIEvent*(eventName: string, body: untyped) =
   ## Dispatches an FFI event to the callback registered via `{libName}_set_event_callback`.
   ## `body` is evaluated lazily — only when a callback is registered.
+  ## `body` may produce a `string` (legacy JSON style) or a `seq[byte]` (CBOR).
   ## Valid only on the FFI thread (i.e., inside {.ffi.} proc bodies and their async closures).
   let ffiState = ffiCurrentCallbackState
   if isNil(ffiState) or isNil(ffiState[].callback):
```
NagyZoltanPeter: `ffiCurrentCallbackState` silently points to the heap allocated lib context's member.
As such it must be guarded, due that xxx_set_event_callback can modify state from another thread.
Note: this threadvar is only set once at lib setup, but at teardown it is not guarantied it cannot outlive the context and issued here before thread is dropped.

ffi/ffi_context.nim
```diff
@@ -83,14 +86,43 @@ template dispatchFFIEvent*(eventName: string, body: untyped) =
     try:
       let event = body
       cast[FFICallBack](ffiState[].callback)(
-        RET_OK, unsafeAddr event[0], cast[csize_t](len(event)), ffiState[].userData
+        RET_OK,
+        cast[ptr cchar](unsafeAddr event[0]),
+        cast[csize_t](len(event)),
+        ffiState[].userData,
       )
     except Exception, CatchableError:
```
NagyZoltanPeter, line 115: claude tells me the cborEncode can raise `Defect` which may leak out from this catch and propagates out of this scope. Worth to check IMO.

NagyZoltanPeter (changes requested):
Good work, although it still stays ffi event processing on the nim thread. Guessing move this away comes in another PR.
I think the accessing of callback store and mutating must be mutexed. Without this it is unsafe.

ffi/ffi_context.nim
Ivansete-status: Thanks! Addressed in db5c3ccebb935839b1cb0eac2aa5565b07407ed1

ffi/codegen/cpp.nim
Ivansete-status: Yes good point! We will address that in https://github.com/logos-messaging/nim-ffi/issues/40

ffi/codegen/cpp.nim
Ivansete-status: Addressed in 2b817de06891ff07872c6cc98469785b7bafbc73

ffi/ffi_context.nim
Ivansete-status: Also addressed by 2b817de06891ff07872c6cc98469785b7bafbc73

ffi/ffi_context.nim
```diff
@@ -47,48 +76,89 @@ var onFFIThread* {.threadvar.}: bool
...
   ## Dispatches an FFI event to the callback registered via `{libName}_set_event_callback`.
   ## `body` is evaluated lazily — only when a callback is registered.
+  ## `body` may produce a `string` (legacy JSON style) or a `seq[byte]` (CBOR).
   ## Valid only on the FFI thread (i.e., inside {.ffi.} proc bodies and their async closures).
   let ffiState = ffiCurrentCallbackState
-  if isNil(ffiState) or isNil(ffiState[].callback):
+  if isNil(ffiState):
+    chronicles.error eventName & " - event callback not set"
+    return
+  let (cbPtr, ud) = snapshotCallback(ffiState[])
```
NagyZoltanPeter, line 106: I believe the whole read of ffiCurrentCallbackState must fall under lock. 
Even if the write is locked this read can catch it midwork... so in theory even if `isNil(ffiState)` is false it can be nil at next read `snapshotCallback(ffiState[])`

NagyZoltanPeter (commented):
Still a lock missing

ffi/ffi_context.nim
Ivansete-status: I think it is addressed in 6f0f44e3319708461dfcbf30bbec4d65dfe04112

ffi/ffi_context.nim
Ivansete-status: In this particular case all reads are made by the FFI thread so, maybe not needed to add explicit lock.

NagyZoltanPeter: Ok, now I see, that additional dereferenced nil check has moved. 
Thanks. You were right 😉

NagyZoltanPeter (approved):

gmelodie (approved):
One issue still bugs me though: between `snapshotCallback` returning and `cb(userdata)` executing, the consumer can re-register and free the object `userdata` pointed at. This could (probably would) result in a use-after-free issue eventually. I don't know how to fix it on the ffi side though.

PR #41, opened 2026-05-22
tests: cover gaps in CBOR type coverage

### LINKED ISSUE #31: Type coverage gaps
The current suite only exercises string, int64, seq[T], Option[T]. Add round-trip cases for:

- integer widths: int8/16/32/64, uint8/16/32/64 (boundary values incl. INT_MIN, UINT64_MAX)
- float32, float64 (incl. NaN, ±inf, subnormals)
- bool
- raw bytes / seq[byte] with embedded NULs
- nested containers: seq[Option[T]], Option[seq[T]], struct-in-struct
- empty seq / empty string
- large payloads (>64 KiB and >1 MiB) to catch allocator/CBOR-size assumptions

gmelodie:
## Summary

Closes #31. Extends `tests/unit/test_serial.nim` with round-trip cases that the original suite (string, int64, seq[T], Option[T]) did not exercise.

Two new suites:
- **CBOR boundaries** — int8/16/32/64 and uint8/16/32/64 at low/high, plus float32/float64 NaN, ±Inf, and smallest subnormal (float32 subnormal compared by raw bit pattern since `math.classify` widens to float64).
- **CBOR round-trips** — `seq[byte]` (incl. empty + embedded NULs), `string` with embedded NULs, three-level struct nesting, `seq[Option[T]]` / `Option[seq[T]]`, `seq[Nested]`, `Option[Nested]`, a `ComplexContainer` with mixed nested fields (populated + all-empty/all-none), and >64 KiB / >1 MiB payloads for both `string` and `seq[byte]`.

Plus a `CBOR void / null sentinel` suite pinning the `CborNullByte` (0xf6) semantics used by `ffi_thread_request` for void replies.

## Test plan

- [x] `nimble test_serial` passes under `--mm:orc` and `--mm:refc` (134 OK assertions across the two MM modes).

### DIFF
--- a/tests/unit/test_serial.nim
+++ b/tests/unit/test_serial.nim
@@ -338,7 +338,7 @@ suite "CBOR boundaries":
     let back = cborDecode(bytes, float64)
     check back.isOk
     check back.value == v
-    check back.value.classify == fcSubnormal
+    check back.value.classify == math.fcSubnormal
 
   test "float32 subnormal (compared by bit pattern)":
     # The smallest positive float32 subnormal has bit pattern 0x00000001.
--- a/tests/unit/test_serial.nim
+++ b/tests/unit/test_serial.nim
@@ -1,4 +1,4 @@
-import std/options
+import std/[math, options]
 import unittest
 import results
 import ffi
@@ -19,6 +19,22 @@ type Color = enum
   cRed
   cGreen
   cBlue
+
+type BytesHolder {.ffi.} = object
+  blob: seq[byte]
+
+type DeepInner {.ffi.} = object
+  tag: string
+  nested: Nested
+
+type ComplexContainer {.ffi.} = object
+  ids: seq[int]
+  names: seq[string]
+  points: seq[Point]
+  maybePoint: Option[Point]
+  maybeNames: Option[seq[string]]
+  flags: seq[Option[int]]
+  blob: seq[byte]
 
 suite "CBOR primitives round-trip":
   test "bool true":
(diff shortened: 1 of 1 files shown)


tests/unit/test_serial.nim
```diff
@@ -242,3 +258,322 @@ suite "cborEncodeShared":
...
+    for v in [low(int16), int16(-1), int16(0), int16(1), high(int16)]:
+      let bytes = cborEncode(v)
+      check cborDecode(bytes, int16).value == v
+
+  test "int32":
+    for v in [low(int32), int32(-1), int32(0), int32(1), high(int32)]:
+      let bytes = cborEncode(v)
+      check cborDecode(bytes, int32).value == v
+
+  test "int64 (INT64_MIN / INT64_MAX)":
```
Ivansete-status: For consistency?
```suggestion
  test "int64":
```

tests/unit/test_serial.nim
```diff
@@ -242,3 +258,322 @@ suite "cborEncodeShared":
...
+    check cborDecode(cborEncode(neg), float32).value == neg
+
+  test "float64 subnormal":
+    # 5e-324 is the smallest positive float64 subnormal (Double.MIN_VALUE).
+    let v = 5e-324
+    let bytes = cborEncode(v)
+    let back = cborDecode(bytes, float64)
+    check back.isOk
+    check back.value == v
+    check back.value.classify == fcSubnormal
```
Ivansete-status: To make it a bit more explicit
```suggestion
    check back.value.classify == math.fcSubnormal
```

Ivansete-status (approved):
LGTM! Thanks for it! 🙌

PR #42, opened 2026-05-23
test(cpp-e2e): add multi-context, cross-library, pipeline, stress tests (#30)

### LINKED ISSUE #30: Add C++ test cases under tests/e2e/cpp/
Add to test_timer_e2e.cpp (or split into focused files as it grows):

- MultiContextIsolation — N independent contexts keep their own state; an error returned on context A does not poison context B; Result/error string is propagated as expected.

- CrossLibrary — load libecho + libmy_timer in the same process from C++, exercise both, ensure no symbol clash and no shared global state. Requires generating a second set of bindings (echoer) — extend the suite's CMakeLists.txt accordingly.

- TriplePipeline — chain three async calls (A → B → C) via .then/futures and assert ordering + payload integrity.

- ThreadedHammer — std::thread-spawned consumers issuing concurrent async calls against one and many contexts; assert no data races / no lost callbacks under -fsanitize=thread.

gmelodie:
## Summary

Implements the four C++ end-to-end test cases requested in issue #30, plus the reviewer's short-lived-thread stress variants. Closes #30.

### New tests (`tests/e2e/cpp/test_timer_e2e.cpp`)

- **`MultiContextIsolation`** — 5 independent `MyTimerCtx` contexts; uses `schedule()` with an empty `JobSpec.name` to force the Nim-side `err("job name must not be empty")`, asserts the string round-trips as a thrown `std::runtime_error`, and verifies the failing context plus all siblings remain usable after the error.
- **`CrossLibrary`** — creates `MyTimerCtx` + `EchoCtx` in the same process, interleaves sync and async calls, asserts each library carries its own state (`timerName="x-timer"` vs `prefix="X-ECHO"`).
- **`TriplePipeline`** — chains three `echoAsync` calls A→B→C inside one outer `std::async`, feeding each `.get()` payload into the next request, asserts ordering and the final `"A->B->C"` payload.
- **`StressShortLivedPerThreadContext` / `StressShortLivedSharedContext`** — reviewer's request: 16/32 threads each fire one request and join; per-thread variant churns the FFI context pool, shared variant stresses multi-producer dispatch into one context.

### Second library (`libecho`)

- `examples/echo/echo.nim`, `echo.nimble` — minimal Nim FFI library with `Echo` type + `shout`/`version` procs.
- New `nimble genbindings_cpp_echo` task; `test_cpp_e2e` and `test_cpp_e2e_sanitized` now regenerate both timer + echo bindings.

### Codegen template changes (required so two nim-ffi headers can coexist in one TU)

- `cbor_helpers.hpp.tpl` and `sync_call_helper.hpp.tpl` wrapped in `NIM_FFI_*_HPP_INCLUDED` guards — otherwise including `my_timer.hpp` + `echo.hpp` in the same translation unit redefines the shared `inline encode_cbor`/`decode_cbor` overloads and the unnamed-namespace `ffi_call_`/`ffi_cb_` helpers.
- `CMakeLists.txt.tpl` — `nim_lib`/`example` CMake targets prefixed with `{{LIB}}_`, and the vendored `tinycbor` static-lib target guarded by `if(NOT TARGET tinycbor)`, so a parent project can `add_subdirectory` two sibling `cpp_bindings/` dirs.
- `tests/e2e/cpp/CMakeLists.txt` updated to `add_subdirectory` both example bindings, link both `*_headers`, and embed both rpaths on Linux/macOS (or stage both DLLs on Windows).

### DIFF
--- a/examples/echo/cpp_bindings/echo.hpp
+++ b/examples/echo/cpp_bindings/echo.hpp
@@ -18,13 +18,8 @@ extern "C" {
 
 // ── encode_cbor overloads (primitives + containers) ─────────────────────
 // Per-struct encode_cbor / decode_cbor are emitted by cpp.nim next to each
-// generated struct. These helpers cover the leaf types and container shapes
-// the struct emitters defer into.
-//
-// Guarded so two nim-ffi headers (e.g. my_timer.hpp + echo.hpp) can be
-// included in the same translation unit — otherwise the second include
-// would redefine these inline overloads. The block ends with the public
-// entry points (encodeCborFFI / decodeCborFFI) before #endif.
+// generated struct; these helpers cover the leaf types they defer into.
+// Guarded so two nim-ffi headers can share a translation unit.
 #ifndef NIM_FFI_CBOR_HELPERS_HPP_INCLUDED
 #define NIM_FFI_CBOR_HELPERS_HPP_INCLUDED
 
@@ -337,14 +332,13 @@ void* echo_create(const uint8_t* req_cbor, size_t req_cbor_len, FFICallback call
 int echo_shout(void* ctx, FFICallback callback, void* user_data, const uint8_t* req_cbor, size_t req_cbor_len);
 int echo_version(void* ctx, FFICallback callback, void* user_data, const uint8_t* req_cbor, size_t req_cbor_len);
 int echo_destroy(void* ctx);
+void echo_set_event_callback(void* ctx, FFICallback callback, void* user_data);
 } // extern "C"
 
 // ============================================================
 // Synchronous call helper
 // ============================================================
-//
-// Guarded so two nim-ffi headers can be included in the same translation
-// unit without redefining ffi_cb_ / ffi_call_ in the unnamed namespace.
+// Guarded so two nim-ffi headers can share a translation unit.
 #ifndef NIM_FFI_SYNC_CALL_HELPER_HPP_INCLUDED
 #define NIM_FFI_SYNC_CALL_HELPER_HPP_INCLUDED
 
--- a/examples/echo/cpp_bindings/echo.hpp
+++ b/examples/echo/cpp_bindings/echo.hpp
@@ -1,0 +1,479 @@
+#pragma once
+#include <string>
+#include <cstdint>
+#include <chrono>
+#include <stdexcept>
+#include <mutex>
+#include <condition_variable>
+#include <memory>
+#include <functional>
+#include <future>
+#include <vector>
+#include <optional>
+#include <type_traits>
+#include <cstring>
+extern "C" {
+#include <tinycbor/cbor.h>
+}
+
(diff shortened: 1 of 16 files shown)


examples/echo/cpp_bindings/echo.hpp
```diff
@@ -0,0 +1,473 @@
...
+}
+
+inline std::vector<std::uint8_t> ffi_call_(std::function<int(FFICallback, void*)> f,
+                                          std::chrono::milliseconds timeout) {
+    auto state = std::make_shared<FFICallState_>();
+    auto* cb_ref = new std::shared_ptr<FFICallState_>(state);
+    const int ret = f(ffi_cb_, cb_ref);
+    if (ret == 2) {
+        delete cb_ref;
+        throw std::runtime_error("RET_MISSING_CALLBACK (internal error)");
```
Ivansete-status, line 381: Maybe we need to revisit the auto-generate code and avoid throwing exceptions.
It seems the `std::excepted` (Result in cpp) is only available in C++23. @NagyZoltanPeter do you think we could enforce C++23 in the auto-generated code?

tests/e2e/cpp/CMakeLists.txt
```diff
@@ -29,32 +33,34 @@ FetchContent_MakeAvailable(googletest)
...
     target_link_options(timer_e2e_tests PRIVATE ${NIM_FFI_SAN_LFLAGS})
 endif()
 
-# The Nim-built shared library has install_name `@rpath/libmy_timer.dylib`
-# (set by `declareLibrary` on macOS for portability). The test binary must
-# therefore know where to find that dylib at load time — embed the build-tree
-# directory of the IMPORTED `my_timer` target as an rpath. On Windows the
-# `my_timer` target is a plain INTERFACE library (no IMPORTED_LOCATION) and
-# we stage the DLL next to the exe via POST_BUILD instead.
+# Nim-built dylibs use `@rpath/lib*.dylib` install_names on macOS, so embed
```
Ivansete-status: I guess that rpath comment also applies to linux?

ffi/internal/ffi_library.nim
```diff
@@ -74,12 +79,26 @@ macro declareLibraryBase*(libraryName: static[string]): untyped =
...
+      ## watchdog spins up a chronos dispatcher.
+      block waitForInit:
+        while true:
+          var expected: int = 0
+          if initState.compareExchange(expected, 1):
+            `nimMainName`()
+            initState.store(2)
+            break waitForInit
+          if initState.load() == 2:
+            break waitForInit
```
Ivansete-status: It seems this can be simplified with the following:
```suggestion
      var expected: int = 0
      if initState.compareExchange(expected, 1):
        `nimMainName`()
        initState.store(2)
      else:
        while initState.load() != 2:
```

Ivansete-status (approved):
LGTM! Thanks for it! 🙌 
Just adding some minot nits that I hope you find useful.
Cheers

Issue #6, opened 2026-01-13
Move user event code to a separate thread


Ivansete-status:
This is a @NagyZoltanPeter 's suggestion. Thanks so much indeed for it and feel free to extend if needed 🥳 

------

## Description

Nowadays, there is a risk that the user code base can interfere the library implementation. For example, in [lmn](https://github.com/logos-messaging/logos-messaging-nim), if the code handling event gets stuck for a while, that can impact the node's health and capability to attend network events.

## Suggested solution

- The library thread will just add an event to a queue, without any blocking or time consuming task. Just enqueue.
- Create a separate thread that will be in charge of attending the library's async events from the event queue.
- On the other hand, we need to modify the current "watchdog" mechanism. Given that we will introduce a new thread (event queue thread,) ideally we should remove the current "watchdog" thread and, instead, have a logic that checks that the library thread is not blocked for a long time (maybe 1 second.)

PR #43, opened 2026-05-26
improve event handling

### LINKED ISSUE #40: Have richer event listener handling
This is inspired by the following question: https://github.com/logos-messaging/nim-ffi/pull/39#discussion_r3288214011

We currently have one `set_event_callback` that is too rigid and doesn't allow to add more listeners later on, or remove event listeners. We need to enhance that.

Ivansete-status:
## Improve event handling

Closes #40 (multi-listener registry + add/remove ABI). Lays the groundwork for #6 (follow-up PR).

## Summary

- Replaces the single-slot `<lib>_set_event_callback` C ABI with a multi-listener registry exposing `<lib>_add_event_listener(ctx, event_name, cb, ud) -> uint64` and `<lib>_remove_event_listener(ctx, id) -> cint`. Each listener gets a stable monotonic id and can be removed independently.
- Registry lives in a new module `ffi/ffi_events.nim`. `dispatchFFIEvent` / `dispatchFFIEventCbor` move there too; `ffi_context.nim` keeps only `FFIContext` + thread lifecycle plus an `eventRegistry: FFIEventRegistry` field.
- `declareLibrary` in `ffi/internal/ffi_library.nim` now emits the two listener-ABI procs. The legacy `_set_event_callback` is removed end-to-end (no Nim helper, no C ABI symbol, no foreign-side caller).

## Foreign-side bindings (C++ and Rust)

The codegen emits a richer, type-safe API per `{.ffiEvent.}`:

### Per-event typed methods

One method per declared event, with the payload type baked in at compile time:

```cpp
// C++
ListenerHandle addOnEchoFiredListener(std::function<void(const EchoEvent&)> handler);
```

```rust
// Rust
pub fn add_on_echo_fired_listener<F>(&self, handler: F) -> ListenerHandle
where F: Fn(&EchoEvent) + Send + Sync + 'static;
```

These are the recommended path — the handler receives the already-decoded payload, no CBOR work in user code.

### Wildcard (catch-all) listener

A single overload that subscribes to **every** event the lib emits, regardless of name. Internally registers with an empty `event_name` against the C ABI; the lib's wildcard trampoline pre-extracts the `eventType` field from the CBOR envelope so the consumer doesn't have to walk CBOR by hand.

```cpp
// C++
ListenerHandle addEventListener(
    std::function<void(int retCode,
                       const std::string& eventId,
                       const std::vector<std::uint8_t>& envelope)> handler);
```

```rust
// Rust
pub fn add_event_listener<F>(&self, handler: F) -> ListenerHandle
where F: Fn(c_int, &str, &[u8]) + Send + Sync + 'static;
```

The handler arguments:

| Arg | Meaning |
|---|---|
| `retCode` / `ret` | FFI return code. `0` = success (envelope is a valid CBOR `EventEnvelope`). Non-zero = error (the bytes are an error string, not a CBOR envelope; `eventId` will be empty). |
| `eventId` / `event_id` | Wire `eventType` string from the envelope (`"on_echo_fired"` for events declared `{.ffiEvent: "on_echo_fired".}`). Empty if `retCode != 0` or the envelope is malformed. |
| `envelope` | Full CBOR envelope bytes (`{ eventType: tstr, payload: <T> }`), suitable to pass to `decodeEventPayload<T>` for the matching event id. |

#### Why a wildcard?

Use cases that the typed per-event methods don't cover cleanly:

- **Diagnostics / logging**: one handler that prints every event the lib emits, without subscribing to each name individually. Doubles as a tracing hook.
- **Generic forwarders**: e.g. forwarding every event to a message bus / IPC channel where decoding by type isn't needed.
- **Future-proofing**: catches events introduced by a newer lib version that the binding wasn't recompiled against.
- **Error channel**: the wildcard sees `retCode != 0` events too. Per-event typed listeners skip those by design (the payload can't be decoded as the registered type), so the wildcard is the place to handle dispatch-time errors.

#### Decoding the payload by type

A helper is emitted alongside the codecs so wildcard handlers can lift the payload into a typed value without rolling their own CBOR walk:

```cpp
// C++
template <class T>
bool decodeEventPayload(const std::vector<std::uint8_t>& envelope, T& out);

// Usage from a wildcard handler:
ctx->addEventListener(
    [&](int retCode, const std::string& eventId, const std::vector<std::uint8_t>& envelope) {
        if (retCode != 0) { /* handle error */ return; }
        if (eventId == "on_echo_fired") {
            EchoEvent evt{};
            if (decodeEventPayload(envelope, evt)) {
                // use evt.message, evt.echoCount, ...
            }
        }
    });
```

```rust
// Rust
pub fn decode_event_payload<T: serde::de::DeserializeOwned>(envelope: &[u8]) -> Result<T, String>;

// Usage:
ctx.add_event_listener(move |ret, event_id, envelope| {
    if ret == 0 && event_id == "on_echo_fired" {
        if let Ok(evt) = decode_event_payload::<EchoEvent>(envelope) {
            // use evt.message, evt.echo_count, ...
        }
    }
});
```

`decodeEventPayload` / `decode_event_payload` understands any `{.ffi.}`-declared type already registered with the binding generator — same codecs the typed methods use under the hood.

### Remove by handle

```cpp
bool removeEventListener(ListenerHandle handle);   // C++
```

```rust
pub fn remove_event_listener(&self, handle: ListenerHandle) -> bool;   // Rust
```

Listeners are owned by the context — `std::unordered_map<uint64_t, std::unique_ptr<ListenerBase>>` in C++, `Mutex<HashMap<u64, Box<dyn Any + Send>>>` in Rust — so the user's closure stays alive until the matching `removeEventListener` call returns.

## Examples

- `examples/timer/cpp_bindings/main.cpp` — section [7] demonstrates typed listener + wildcard with `eventId` dispatch + `decodeEventPayload<EchoEvent>` + `removeEventListener`.
- `examples/timer/rust_bindings/examples/main.rs` — sync example.
- `examples/timer/rust_bindings/examples/tokio_main.rs` — async example via `#[tokio::main]` (Cargo.toml now ships a `[dev-dependencies]` block with `rt-multi-thread` + `macros` so `examples/` can use the macro without bloating the library's runtime deps).

## Tests

- `tests/unit/test_event_listener.nim` — **13 new tests** covering mutation, monotonic ids, snapshot semantics, multi-listener dispatch, wildcard, re-entrant add (does not fire on current dispatch), re-entrant remove (in-flight snapshot still delivers once), and unknown-id removal.
- `tests/unit/test_event_dispatch.nim` — migrated to `addEventListener`; concurrent stress suite gated to non-refc with a comment (refc's per-thread GC heaps can't safely grow a `seq[T]` across threads even under a lock).
- `tests/e2e/cpp/test_timer_e2e.cpp` — 3 new tests: `MultipleTypedListenersAllFire`, `RemoveEventListenerStopsDelivery`, `WildcardListenerReceivesEventIdAndDecodesPayload`. Total 14/14.

## Other small changes

- "No listener registered" log demoted from `error` to `debug` — foreign code legitimately not subscribing to every event isn't an error.
- `chronicles.error` on the FFI thread now reports only actual programmer errors (registry not bound to the thread).

## Known follow-ups (out of scope for this PR)

1. **Snapshot/invoke UAF window** — second concern in #40. `snapshotListeners` copies the listener seq under lock then iterates outside, so a foreign `removeEventListener` can free the listener's heap object before `cb(userData)` runs. Will be addressed in the #6 follow-up by switching to a recursive registry mutex held during invocation (cheap once dispatch lives on the dedicated event thread).
2. **#6** — move dispatch off the FFI thread (bounded MPSC queue + dedicated event thread + system events `__queue_near_full__` / `__queue_full__` / `__not_responding__`); fold the standalone watchdog into the new event thread.

## Test plan

- [x] `nimble test` — all suites green under `--mm:orc` and `--mm:refc`.
- [x] `nimble check_bindings_cpp` / `nimble check_bindings_rust` — checked-in bindings reproduce.
- [x] `nimble test_cpp_e2e` — 14/14 (including the 3 new event tests).
- [x] `cargo run --example main` / `cargo run --example tokio_main` from `examples/timer/rust_bindings`.
- [x] C++ `example` binary prints sections [1]–[7] cleanly.
- [ ] Optional: `NIM_FFI_SAN=asan-ubsan` and `=tsan` sanitizer passes (worth running given the #40 race history).

### DIFF
--- a/ffi/internal/ffi_library.nim
+++ b/ffi/internal/ffi_library.nim
@@ -92,47 +92,85 @@ macro declareLibraryBase*(libraryName: static[string]): untyped =
   return res
 
 macro declareLibrary*(libraryName: static[string], libType: untyped): untyped =
-  ## Declares a library with the given name and automatically generates
-  ## `{libraryName}_set_event_callback`, a C-exported function that stores the
-  ## caller's event callback on the FFIContext.
+  ## Declares a library with the given name and emits the C-exported event
+  ## ABI on its `FFIContext`:
   ##
-  ## `libType` is the Nim type of the main library object (e.g. `Waku`). It is used
-  ## to type the `ctx: ptr FFIContext[libType]` parameter of the generated
-  ## `{libraryName}_set_event_callback` proc.
+  ## - `{libraryName}_add_event_listener(ctx, event_name, cb, ud) -> uint64`
+  ##   — registers `cb` for `event_name` and returns its stable id. An
+  ##   empty `event_name` subscribes `cb` to *every* event (catch-all).
+  ## - `{libraryName}_remove_event_listener(ctx, id) -> cint` — returns 0 on
+  ##   success, non-zero if no listener with that id exists.
+  ##
+  ## `libType` is the Nim type of the main library object, used to type
+  ## the `ctx: ptr FFIContext[libType]` parameter. See
+  ## `examples/timer/timer.nim` for a working call site.
   var stmts = newStmtList()
 
   # Emit the base bootstrap (pragmas, linker flags, NimMain, initializeLibrary)
   stmts.add(newCall(ident("declareLibraryBase"), newStrLitNode(libraryName)))
 
-  let funcName = libraryName & "_set_event_callback"
-  let funcIdent = ident(funcName)
-  let errorMsg = "error: invalid context in " & funcName
-
   let ctxType = nnkPtrTy.newTree(nnkBracketExpr.newTree(ident("FFIContext"), libType))
-
-  let procBody = quote:
-    if isNil(ctx):
-      echo `errorMsg`
-      return
-    setCallback(ctx[].callbackState, cast[pointer](callback), userData)
-
-  let procNode = newProc(
-    name = funcIdent,
-    params = @[
-      newEmptyNode(),
-      newIdentDefs(ident("ctx"), ctxType),
-      newIdentDefs(ident("callback"), ident("FFICallBack")),
-      newIdentDefs(ident("userData"), ident("pointer")),
-    ],
-    body = procBody,
-    pragmas = newTree(
-      nnkPragma,
-      ident("dynlib"),
-      ident("exportc"),
-      ident("cdecl"),
-      newTree(nnkExprColonExpr, ident("raises"), newTree(nnkBracket)),
-    ),
+  let cdeclExportPragma = newTree(
+    nnkPragma,
+    ident("dynlib"),
+    ident("exportc"),
+    ident("cdecl"),
+    newTree(nnkExprColonExpr, ident("raises"), newTree(nnkBracket)),
   )
 
-  stmts.add(procNode)
+  # {libraryName}_add_event_listener
+  let addName = libraryName & "_add_event_listener"
+  let addErr = "error: invalid context in " & addName
+  let addBody = quote:
+    var ret: uint64 = 0
+    if isNil(ctx):
+      echo `addErr`
+      return ret
+    let evtName = if eventName.isNil(): "" else: $eventName
+    ret = addEventListener(ctx[].eventRegistry, evtName, callback, userData)
+    return ret
+
+  stmts.add(
+    newProc(
+      name = ident(addName),
+      params = @[
+        ident("uint64"),
+        newIdentDefs(ident("ctx"), ctxType),
+        newIdentDefs(ident("eventName"), ident("cstring")),
+        newIdentDefs(ident("callback"), ident("FFICallBack")),
+        newIdentDefs(ident("userData"), ident("pointer")),
+      ],
+      body = addBody,
+      pragmas = cdeclExportPragma,
+    )
+  )
+
+  # --- {libraryName}_remove_event_listener --------------------------------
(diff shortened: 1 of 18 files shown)


ffi/internal/ffi_library.nim
```diff
@@ -92,47 +92,85 @@ macro declareLibraryBase*(libraryName: static[string]): untyped =
...
-  let errorMsg = "error: invalid context in " & funcName
-
   let ctxType = nnkPtrTy.newTree(nnkBracketExpr.newTree(ident("FFIContext"), libType))
+  let cdeclExportPragma = newTree(
+    nnkPragma,
+    ident("dynlib"),
+    ident("exportc"),
+    ident("cdecl"),
+    newTree(nnkExprColonExpr, ident("raises"), newTree(nnkBracket)),
+  )
```
gmelodie, line 119: Maybe we could add comments with the expected code generated by these calls, much like what is in line 121: `{libraryName}_add_event_listener`. This makes me think that we could have more readable code that would resemble pseudocode better...

Ivansete-status:
Sorry for the spam
As per @gmelodie suggestion, I will split this PR in smaller ones

PR #45, opened 2026-05-26
add FFIEventRegistry: multi-listener data structure for FFI events

Ivansete-status:
## Summary

First slice of the multi-listener event refactor (issue #40), splitting the monolithic #43 into reviewable steps.

This PR introduces only the **data structure** — no FFIContext integration, no dispatch, no ABI changes. Subsequent PRs will:

1. Wire dispatch templates over the registry; emit the new `_add_event_listener` / `_remove_event_listener` C ABI; keep `_set_event_callback` as a thin compat shim that calls `addEventListener("", cb, ud)` so existing C++/Rust bindings continue to link unchanged.
2. Add dispatch-level tests + migrate the existing `test_event_dispatch.nim` to the new API.
3. C++ codegen for typed/wildcard listener methods + `decodeEventPayload<T>` helper; regen `my_timer.hpp`.
4. C++ example + e2e test migration.
5. Rust codegen + bindings + examples (in two PRs to stay under the line budget).
6. Drop the `_set_event_callback` compat shim end-to-end.

Total split: 6–7 PRs, each under ~400 lines changed.

## What's in this PR

- `ffi/ffi_events.nim` (new) — `FFIEventRegistry` type, lifecycle (`init`/`deinit`), mutation (`addEventListener` / `removeEventListener` / `removeAllEventListeners`), and `snapshotListeners`. Lock-guarded; monotonic ids; empty event name `""` is the wildcard channel.
- `ffi.nim` — re-export the new module.
- `tests/unit/test_event_listener.nim` (new) — 7 tests covering monotonic id assignment, nil-callback rejection, unknown-id removal, removal across per-event seq and wildcard, snapshot includes wildcard, snapshot-is-a-copy semantics, and removeAll.

Dispatch-level tests (multi-listener fan-out, re-entrancy, wildcard delivery) land alongside the dispatch wiring in PR 2.

## Why review this in isolation

The registry's mutation/snapshot contract is the load-bearing piece that everything downstream depends on. Reviewing it without also reasoning about FFIContext lifecycle or C ABI shape keeps the scope tight.

## Test plan

- [x] ``nimble test`` — all suites green under ``--mm:orc`` and ``--mm:refc`` (7 new tests in ``test_event_listener``, no regressions elsewhere).

### DIFF
--- a/ffi/ffi_events.nim
+++ b/ffi/ffi_events.nim
@@ -1,0 +1,135 @@
+## Multi-listener registry primitive for FFI library-initiated events.
+##
+## Each event name maps to a `seq` of listeners; the empty event name `""`
+## is the wildcard channel and receives every dispatched event in addition
+## to its own per-name subscribers.
+##
+## This module ships the data structure only. Dispatch templates that
+## consume the registry, plus the FFIContext wiring that exposes the
+## registry to {.ffi.}-generated code, land in follow-up PRs. The
+## registry is thread-safe via an embedded `Lock`; the future dispatch
+## path acquires the lock only long enough to copy out the listener
+## slice for the event being dispatched, so re-entrant add/remove from
+## within a handler is deadlock-free by construction.
+
+import std/[locks, tables]
+import ./ffi_types
+
+# ---------------------------------------------------------------------------
+# Registry types
+# ---------------------------------------------------------------------------
+
+type
+  FFIEventListener* = object
+    id*: uint64
+    callback*: FFICallBack
+    userData*: pointer
+
+  FFIEventRegistry* = object
+    ## Per-context multi-listener registry. `lock` guards every mutation;
+    ## readers (dispatch path) acquire it only long enough to copy out the
+    ## listener slice for the event being dispatched.
+    lock*: Lock
+    nextId*: uint64
+      ## Monotonic id source. 0 is reserved as "invalid"; ids start at 1.
+    byEvent*: Table[string, seq[FFIEventListener]]
+    wildcard*: seq[FFIEventListener]
+
+const WildcardEventName* = ""
+  ## Empty string registers a wildcard listener that receives every event.
+
+# ---------------------------------------------------------------------------
+# Registry lifecycle and mutation
+# ---------------------------------------------------------------------------
+
+proc initEventRegistry*(reg: var FFIEventRegistry) =
+  reg.lock.initLock()
+  reg.nextId = 0'u64
+  reg.byEvent = initTable[string, seq[FFIEventListener]]()
+  reg.wildcard.setLen(0)
+
+proc deinitEventRegistry*(reg: var FFIEventRegistry) =
+  reg.lock.deinitLock()
+  reg.byEvent.clear()
+  reg.wildcard.setLen(0)
+
+proc addEventListener*(
+    reg: var FFIEventRegistry,
+    eventName: string,
+    callback: FFICallBack,
+    userData: pointer,
+): uint64 {.raises: [].} =
+  ## Registers `callback` for `eventName` and returns the listener's stable
+  ## id (always non-zero on success). `eventName == ""` registers a wildcard
+  ## listener that receives every dispatched event. Returns 0 if `callback`
+  ## is nil — the only documented failure mode.
+  var assigned: uint64 = 0
+  if callback.isNil():
+    return assigned
+
+  withLock reg.lock:
+    reg.nextId += 1
+    assigned = reg.nextId
+    let listener =
+      FFIEventListener(id: assigned, callback: callback, userData: userData)
+    if eventName.len == 0:
+      reg.wildcard.add(listener)
+    else:
+      # `mgetOrPut` lets us avoid a `[]` lookup that the effect tracker
+      # would otherwise see as raising `KeyError`.
+      reg.byEvent.mgetOrPut(eventName, @[]).add(listener)
+  return assigned
+
+proc removeEventListener*(reg: var FFIEventRegistry, id: uint64): bool {.raises: [].} =
+  ## Removes the listener with `id`. Returns true on success, false if no
+  ## listener with that id exists. Safe to call from inside a dispatch:
+  ## the in-flight snapshot still delivers exactly once to the listener
+  ## being removed.
+  var removed = false
+  if id == 0'u64:
+    return removed
+
+  withLock reg.lock:
+    for i in 0 ..< reg.wildcard.len:
+      if reg.wildcard[i].id == id:
+        reg.wildcard.delete(i)
+        removed = true
+        break
+    if not removed:
+      for listeners in reg.byEvent.mvalues:
+        var idx = -1
+        for i in 0 ..< listeners.len:
+          if listeners[i].id == id:
+            idx = i
+            break
+        if idx >= 0:
+          listeners.delete(idx)
+          removed = true
+          break
+  return removed
+
+proc removeAllEventListeners*(reg: var FFIEventRegistry) {.raises: [].} =
+  ## Drops every registered listener (per-event and wildcard). Does not
+  ## reset the listener-id counter — subsequent `addEventListener` calls
+  ## still return strictly increasing ids.
+  withLock reg.lock:
+    reg.wildcard.setLen(0)
+    reg.byEvent.clear()
+
+proc snapshotListeners*(
+    reg: var FFIEventRegistry, eventName: string
+): seq[FFIEventListener] {.raises: [].} =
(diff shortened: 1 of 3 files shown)


ffi/ffi_events.nim
```diff
@@ -0,0 +1,135 @@
...
+        var idx = -1
+        for i in 0 ..< listeners.len:
+          if listeners[i].id == id:
+            idx = i
+            break
+        if idx >= 0:
+          listeners.delete(idx)
+          removed = true
+          break
+  return removed
```
fcecin, line 125: Empty keys can be pruned here? Example:

```
  -      for listeners in reg.byEvent.mvalues:
  +      var emptyKey = ""
  +      var prune = false
  +      for key, listeners in reg.byEvent.mpairs:
           var idx = -1
           for i in 0 ..< listeners.len:
             if listeners[i].id == id:
               idx = i
               break
           if idx >= 0:
             listeners.delete(idx)
             removed = true
  +          if listeners.len == 0:
  +            emptyKey = key
  +            prune = true
             break
  +      if prune:
  +        reg.byEvent.del(emptyKey)
```

fcecin (commented):
LGTM

fcecin (approved):

ffi/ffi_events.nim
```diff
@@ -0,0 +1,135 @@
...
+
+# ---------------------------------------------------------------------------
+# Registry lifecycle and mutation
+# ---------------------------------------------------------------------------
+
+proc initEventRegistry*(reg: var FFIEventRegistry) =
+  reg.lock.initLock()
+  reg.nextId = 0'u64
+  reg.byEvent = initTable[string, seq[FFIEventListener]]()
+  reg.wildcard.setLen(0)
```
gmelodie, line 53: Any reason this is an `init` that receives a `var` instead of a `new` that returns the registry? If we're to keep it this way, does it make sense to assume that other fields might need to be protected from the get go? i.e. we'd need to do this

```suggestion
proc initEventRegistry*(reg: var FFIEventRegistry) =
  reg.lock.initLock()
  withLock reg.lock:
    reg.nextId = 0'u64
    reg.byEvent = initTable[string, seq[FFIEventListener]]()
    reg.wildcard.setLen(0)
```

ffi/ffi_events.nim
```diff
@@ -0,0 +1,135 @@
...
+
+proc deinitEventRegistry*(reg: var FFIEventRegistry) =
+  reg.lock.deinitLock()
+  reg.byEvent.clear()
+  reg.wildcard.setLen(0)
```
gmelodie, line 63: Similarly as above, would it make sense to deinit the lock at the end?

```suggestion
proc deinitEventRegistry*(reg: var FFIEventRegistry) =
  withLock reg.lock:
    reg.byEvent.clear()
    reg.wildcard.setLen(0)
  reg.lock.deinitLock()
```

gmelodie: Also it might be just me but I've always disliked the name deinit. I mean... de-init???? what? kek

ffi/ffi_events.nim
```diff
@@ -0,0 +1,135 @@
...
+    userData: pointer,
+): uint64 {.raises: [].} =
+  ## Registers `callback` for `eventName` and returns the listener's stable
+  ## id (always non-zero on success). `eventName == ""` registers a wildcard
+  ## listener that receives every dispatched event. Returns 0 if `callback`
+  ## is nil — the only documented failure mode.
+  var assigned: uint64 = 0
+  if callback.isNil():
+    return assigned
+
```
gmelodie: nit
```suggestion
  if callback.isNil():
    return 0
    
  var assigned: uint64
```

ffi/ffi_events.nim
```diff
@@ -0,0 +1,135 @@
...
+  withLock reg.lock:
+    reg.nextId += 1
```
gmelodie: nit
```suggestion
    reg.nextId.inc()
```

ffi/ffi_events.nim
```diff
@@ -0,0 +1,135 @@
...
+    callback: FFICallBack,
+    userData: pointer,
+): uint64 {.raises: [].} =
+  ## Registers `callback` for `eventName` and returns the listener's stable
+  ## id (always non-zero on success). `eventName == ""` registers a wildcard
+  ## listener that receives every dispatched event. Returns 0 if `callback`
+  ## is nil — the only documented failure mode.
+  var assigned: uint64 = 0
+  if callback.isNil():
+    return assigned
```
gmelodie: Should we check `eventName != ""` as well?

ffi/ffi_events.nim
```diff
@@ -0,0 +1,135 @@
...
+  withLock reg.lock:
+    reg.nextId += 1
+    assigned = reg.nextId
+    let listener =
+      FFIEventListener(id: assigned, callback: callback, userData: userData)
+    if eventName.len == 0:
+      reg.wildcard.add(listener)
+    else:
+      # `mgetOrPut` lets us avoid a `[]` lookup that the effect tracker
+      # would otherwise see as raising `KeyError`.
```
```suggestion
```

ffi/ffi_events.nim
```diff
@@ -0,0 +1,135 @@
...
+  return assigned
+
+proc removeEventListener*(reg: var FFIEventRegistry, id: uint64): bool {.raises: [].} =
+  ## Removes the listener with `id`. Returns true on success, false if no
+  ## listener with that id exists. Safe to call from inside a dispatch:
+  ## the in-flight snapshot still delivers exactly once to the listener
+  ## being removed.
+  var removed = false
+  if id == 0'u64:
+    return removed
```
gmelodie: same nit as above
```suggestion
  if id == 0:
    return removed

  var removed = false
```

ffi/ffi_events.nim
```diff
@@ -0,0 +1,135 @@
...
+      for listeners in reg.byEvent.mvalues:
+        var idx = -1
+        for i in 0 ..< listeners.len:
+          if listeners[i].id == id:
+            idx = i
+            break
+        if idx >= 0:
+          listeners.delete(idx)
+          removed = true
+          break
```
gmelodie, line 122: (don't accept this suggestion, cuz the template needs to be moved out of this proc)
```suggestion
template removeById(s: var seq[Listener], target): bool =
  var found = false
  for i, x in s:
    if x.id == target:
      s.delete(i)
      found = true
      break
  found

withLock reg.lock:
  removed = reg.wildcard.removeById(id)
  if not removed:
    for listeners in reg.byEvent.mvalues:
      if listeners.removeById(id):
        removed = true
        break
```

gmelodie (approved):
Overall comment: make docstrings less verbose

Loving thiss!

ffi/ffi_events.nim
Ivansete-status: Addressed in 726f87074736227bc3ad3a416e7ec782454069ce

ffi/ffi_events.nim
Ivansete-status: The FFIEventRegistry should be passed as ref so that the stack state is preserved after invoking it. Is a design thing for `initLock` too.
This should be called once by the thread owner. Added a comment here: 310483d2069bfea402b960c9d1eff6d929def539

ffi/ffi_events.nim
Ivansete-status: Added a comment here to make it explicit who should use it: 4f7978b8633f7750bf621b47e160b2b2c85bca14
The `deinit` name comes from the standard library in locks.nim (deinitLock) and just to follow same naming convention.

ffi/ffi_events.nim
Ivansete-status: Addressed in a30392e84e9a03cea38d35b86c6f208431ea5018

ffi/ffi_events.nim
Ivansete-status: eventName == "" is a wildcard thing that means: _add listener to all possible events in the lib_. That's mentioned in the comment above.

ffi/ffi_events.nim
Ivansete-status: Addressed in c5cda8b94fe0f00b2a4a45f78de50a1637b0c1d4

ffi/ffi_events.nim
Ivansete-status: Addressed in 91713dc5a44b466be6a510419e042be7407e80d7

ffi/ffi_events.nim
Ivansete-status: Addressed in 21df96dee6265fbf5b1cd32a8bbbfc1b076ceb33

PR #46, opened 2026-05-26
chore: avoid throwing exceptions in C++ bindings

### LINKED ISSUE #44: Avoid throwing exceptions (needs C++ 23 for Results)
From https://github.com/logos-messaging/nim-ffi/pull/42#discussion_r3300206487:
> Maybe we need to revisit the auto-generate code and avoid throwing exceptions.
It seems the std::excepted (Result in cpp) is only available in C++23. @NagyZoltanPeter do you think we could enforce C++23 in the auto-generated code?

gmelodie:
## Summary

Closes #44. The auto-generated C++ bindings used to surface every failure as a
thrown `std::runtime_error` (FFI timeout, missing callback, CBOR encode/decode
errors, non-numeric ctx address). The original plan was to adopt
`std::expected`, but that requires C++23. Instead, this PR embeds a small
`Result<T>` type directly in the generated header — mirroring the Nim side's
`Result[T, string]` — so the bindings compile on older standards and never throw.

Every fallible entry point now returns a `Result<T>`; callers branch on
`isOk()`/`isErr()` (or the explicit `bool` conversion) and read
`value()`/`error()`.

## Changes

- **New `Result<T>` template** (`ffi/codegen/templates/cpp/result.hpp.tpl`):
  `Result<T>` + `Result<void>` specialization, include-guarded. Emitted right
  after the prelude so the CBOR and sync-call helpers can use it.
- **Helpers no longer throw**:
  - `cbor_helpers.hpp.tpl`: `encodeCborFFI` → `Result<std::vector<uint8_t>>`,
    `decodeCborFFI<T>` → `Result<T>`.
  - `sync_call_helper.hpp.tpl`: `ffi_call_` → `Result<std::vector<uint8_t>>`
    (timeout / `RET_MISSING_CALLBACK` / error payload are returned, not thrown).
  - `header_prelude.hpp.tpl`: drop `<stdexcept>`, add `<charconv>`.
- **Codegen** (`ffi/codegen/cpp.nim`):
  - `create()` returns `Result<std::unique_ptr<Ctx>>`, propagates
    encode/call/decode errors, and parses the ctx address with
    `std::from_chars` instead of the exception-throwing `std::stoull`.
  - Instance methods return `Result<RetType>`; `*Async` wrappers return
    `std::future<Result<RetType>>`; `createAsync` returns
    `std::future<Result<std::unique_ptr<Ctx>>>`.
- **Regenerated bindings + updated consumers**: `my_timer.hpp`, `echo.hpp`,
  `main.cpp`, and `test_timer_e2e.cpp` now branch on `Result`. A `mustOk`
  helper unwraps in single-threaded tests; the threaded stress tests check
  `Result` directly via their atomic counters (gtest macros aren't thread-safe).

The event-trampoline path is untouched — it already used the lower-level
`decode_cbor` (returns `CborError`), not the throwing entry points.

## Breaking change

This changes the generated C++ API: calls now yield `Result<T>` rather than the
value directly. Existing C++ consumers must adapt (see the updated `main.cpp`
and e2e tests for the pattern).

### DIFF
--- a/examples/echo/cpp_bindings/echo.hpp
+++ b/examples/echo/cpp_bindings/echo.hpp
@@ -2,7 +2,7 @@
 #include <string>
 #include <cstdint>
 #include <chrono>
-#include <stdexcept>
+#include <charconv>
 #include <mutex>
 #include <condition_variable>
 #include <memory>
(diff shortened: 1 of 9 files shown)


examples/echo/cpp_bindings/echo.hpp
```diff
@@ -16,6 +16,69 @@ extern "C" {
...
+    }
+    bool isOk() const { return value_.has_value(); }
+    bool isErr() const { return !value_.has_value(); }
+    explicit operator bool() const { return isOk(); }
+    const T& value() const { return *value_; }
+    T& value() { return *value_; }
+    const T& operator*() const { return *value_; }
+    const T* operator->() const { return &*value_; }
+    T&& take() { return std::move(*value_); }
+    const std::string& error() const { return error_; }
```
Ivansete-status: It is interesting to add assert that can help the dev detect possible bugs in debug. That would need a change in result.hpp.tpl

```suggestion
const T& value() const         { assert(value_.has_value() && "Result::value() called on err Result — check isOk() first"); return *value_; }
    T& value()                     { assert(value_.has_value() && "Result::value() called on err Result — check isOk() first"); return *value_; }
    const T& operator*() const     { assert(value_.has_value() && "Result::operator*() called on err Result — check isOk() first"); return *value_; }
    const T* operator->() const    { assert(value_.has_value() && "Result::operator->() called on err Result — check isOk() first"); return &*value_; }
    T&& take()                     { assert(value_.has_value() && "Result::take() called on err Result — check isOk() first"); return std::move(*value_); }
    const std::string& error() const { assert(!value_.has_value() && "Result::error() called on ok Result — check isErr() first"); return error_; }
```

examples/timer/cpp_bindings/my_timer.hpp
```diff
@@ -16,6 +16,69 @@ extern "C" {
...
 
+// ============================================================
+// Result<T> — exception-free error channel
+// ============================================================
+// The generated bindings never throw: every fallible entry point (create,
+// instance methods, and their *Async futures) returns a Result<T>. Callers
+// branch on isOk()/isErr() (or the explicit bool conversion) and read
+// value()/error(). This mirrors the Nim side's Result[T, string] and keeps
+// us off C++23's std::expected. Guarded so two nim-ffi headers can share a
+// translation unit.
```
Ivansete-status:
```suggestion
// us off C++23's std::expected.
```

tests/e2e/cpp/test_timer_e2e.cpp
```diff
@@ -23,8 +28,20 @@
 
 namespace {
 
+// Unwrap a Result<T> in a single-threaded test context. On error it records a
+// non-fatal gtest failure and returns a default-constructed T so the caller
+// can keep going (subsequent expectations will fail loudly).
+template <typename T>
+T mustOk(Result<T> r) {
+    if (r.isErr()) {
+        ADD_FAILURE() << "unexpected FFI error: " << r.error();
```
Ivansete-status: Maybe?
```suggestion
        ADD_FAILURE() << "unexpected FFI error: " << r.error() << " line: " << __LINE__;
```

Ivansete-status (approved):
LGTM! Thanks for it! 🙌 
Just minor nits

PR #47, opened 2026-05-26
fix use-after-free concern

Ivansete-status:
## Close the snapshot/invoke UAF window in event dispatch

Addresses the second concern in #40 — the use-after-free flagged in
[PR #39 review #4356915554](https://github.com/logos-messaging/nim-ffi/pull/39#pullrequestreview-4356915554).

### Problem

`callEventCallback`, `dispatchFFIEvent`, and `dispatchFFIEventCbor`
snapshotted `(callback, userData)` under `callbackState.lock`, then
released the lock and invoked the callback outside it. A foreign-thread
`setCallback` in that window swapped the pair and let the foreign-side
binding free the old `userData`'s backing object — while the in-flight
dispatch still held a pointer to it. UAF on invocation.

### Fix

Each dispatch template now holds `callbackState.lock` for the entire
snapshot + invocation. Concurrent `setCallback` from a foreign thread
blocks until the in-flight callback returns, so the binding cannot free
the heap object behind `userData` until we're done with it.

### Test

New regression in `tests/unit/test_event_dispatch.nim`:
- Worker thread directly drives `dispatchFFIEvent` against a
  `FFICallbackState` whose callback sleeps 15ms after signalling entry.
- Main thread waits for the entry signal, then calls `setCallback`.
- Asserts that by the time `setCallback` returns, the in-flight
  callback has finished (proving the lock blocked the swap).

The test deliberately doesn't use the FFI thread's request channel:
request handlers run synchronously on the FFI thread before
`reqReceivedSignal` fires, so the main thread would never observe the
in-flight state through that path.

### Verification

- `nimble test` — all suites green under `--mm:orc` and `--mm:refc`.
- Diff: 140 additions, 50 deletions across 2 files (190 lines).

### DIFF
--- a/ffi/ffi_context.nim
+++ b/ffi/ffi_context.nim
@@ -83,8 +83,7 @@ template callEventCallback*(ctx: ptr FFIContext, eventName: string, body: untype
   ## concurrent `setCallback` from a foreign thread blocks until the
   ## in-flight callback returns. Without this, the foreign-side binding
   ## could free the object `userData` points at between the snapshot and
-  ## the invocation, causing a use-after-free — issue #40 / PR #39
-  ## review #4356915554.
+  ## the invocation, causing a use-after-free.
   withLock ctx[].callbackState.lock:
     let cbPtr = ctx[].callbackState.callback
     let ud = ctx[].callbackState.userData
--- a/ffi/ffi_context.nim
+++ b/ffi/ffi_context.nim
@@ -78,43 +78,55 @@ const git_version* {.strdefine.} = "n/a"
 template callEventCallback*(ctx: ptr FFIContext, eventName: string, body: untyped) =
   ## `body` may evaluate to a `string` or a `seq[byte]` — the cast to
   ## `ptr cchar` accepts both `ptr char` and `ptr byte` source pointers.
-  let (cbPtr, ud) = snapshotCallback(ctx[].callbackState)
-  if isNil(cbPtr):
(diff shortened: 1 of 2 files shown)


ffi/ffi_context.nim
```diff
@@ -78,43 +78,55 @@ const git_version* {.strdefine.} = "n/a"
...
-        "Exception " & eventName & " when calling 'eventCallBack': " &
-        getCurrentExceptionMsg()
-      cb(RET_ERR, cast[ptr cchar](unsafeAddr msg[0]), cast[csize_t](len(msg)), ud)
+  ##
+  ## Holds `callbackState.lock` for the snapshot + invocation so that a
+  ## concurrent `setCallback` from a foreign thread blocks until the
+  ## in-flight callback returns. Without this, the foreign-side binding
+  ## could free the object `userData` points at between the snapshot and
+  ## the invocation, causing a use-after-free — issue #40 / PR #39
+  ## review #4356915554.
```
```suggestion
  ## the invocation, causing a use-after-free.
```

gmelodie (approved):
Very cool. Does this mean we can remove `snapshotCallback`? Or is it being used somewhere still?

Ivansete-status:
> Very cool. Does this mean we can remove `snapshotCallback`? Or is it being used somewhere still?

This is part of a PRs chain. This cleanup will happen eventually in upcoming PRs. All based on the closed #43

PR #49, opened 2026-05-26
Event registry wiring

Ivansete-status:
Move event dispatch off the single-callback `FFICallbackState` onto the multi-listener `FFIEventRegistry` introduced in #45.

- Replace `callbackState` field on `FFIContext` with `eventRegistry`; the `_set_event_callback` C ABI usage is kept (now backed by `removeAllEventListeners` + `addEventListener(WildcardEventName, …)`) but will be modified in next PRs.

- Move `dispatchFFIEvent` / `dispatchFFIEventCbor` / `EventEnvelope` and the new `ffiCurrentEventRegistry` threadvar into `ffi_events.nim`.

- `onNotResponding` rewritten to dispatch through the registry under the same lock-during-invocation contract.

- `FFIContextPool.destroyFFIContext` now calls `deinitEventRegistry` before returning the slot so the registry's GC-managed `Table` / `seq` storage is freed on the owning thread (otherwise the next slot user crashes inside `initTable` trying to dealloc the previous thread's heap).

- `test_event_dispatch` migrated from `setCallback` to `addEventListener`; refc stress suite gated since cross-thread seq mutation isn't safe under refc's per-thread GC heaps.

Issue: #40


fcecin (approved):
LGTM

ffi/ffi_context.nim
```diff
@@ -242,8 +101,29 @@ proc `$`(event: JsonNotRespondingEvent): string =
...
-    $JsonNotRespondingEvent.init()
+  ## Shim: still emits the legacy JSON payload through the registry, so
+  ## existing foreign consumers see no wire-shape change. A follow-up
+  ## PR replaces this with a CBOR `NotRespondingEvent`.
+  ## Mirrors the dispatch templates' lock-during-invocation contract
+  ## (see `ffi_events.nim`).
+  withLock ctx[].eventRegistry.lock:
+    var snap: seq[FFIEventListener] = @[]
+    for l in ctx[].eventRegistry.byEvent.getOrDefault("onNotResponding"):
+      snap.add(l)
```
```suggestion
    ctx[].eventRegistry.byEvent.getOrDefault("onNotResponding").mapIt(snap.add(it))
```

ffi/ffi_context.nim
```diff
@@ -242,8 +101,29 @@ proc `$`(event: JsonNotRespondingEvent): string =
...
+    for l in ctx[].eventRegistry.wildcard:
+      snap.add(l)
```
gmelodie: Same as above

ffi/ffi_events.nim
```diff
@@ -58,9 +74,14 @@ proc deinitEventRegistry*(reg: var FFIEventRegistry) =
...
+  ##
+  ## Resets the GC-managed fields to default so `FFIContextPool`'s
+  ## slot reuse on a *different* thread doesn't trigger Nim's hidden
+  ## assignment destructor against this thread's heap allocations.
   reg.lock.deinitLock()
-  reg.byEvent.clear()
-  reg.wildcard.setLen(0)
+  reg.byEvent = default(Table[string, seq[FFIEventListener]])
+  reg.wildcard = @[]
+  reg.nextId = 0'u64
```
gmelodie, line 84: Not sure if we need to do the following or not
```nim
withLock reg.lock:
  reg.byEvent = default(Table[string, seq[FFIEventListener]])
  reg.wildcard = @[]
  reg.nextId = 0'u64
reg.lock.deinitLock
```

ffi/ffi_events.nim
```diff
@@ -85,6 +106,8 @@ proc addEventListener*(
     if eventName.len == 0:
       reg.wildcard.add(listener)
     else:
+      # `mgetOrPut` lets us avoid a `[]` lookup that the effect tracker
+      # would otherwise see as raising `KeyError`.
```
```suggestion
```

ffi/ffi_events.nim
```diff
@@ -149,3 +172,105 @@ proc snapshotListeners*(
...
+  ## Valid only on the FFI thread (where `ffiCurrentEventRegistry` is
+  ## set). Holds `reg.lock` for the entire snapshot + invocation so a
+  ## concurrent `removeEventListener` from a foreign thread blocks until
+  ## dispatch returns — closes the UAF window in #40 / PR #39 review
+  ## #4356915554. Handlers must not call addEventListener /
+  ## removeEventListener on the same registry (would self-deadlock).
+  let regPtr = ffiCurrentEventRegistry
+  if regPtr.isNil():
+    chronicles.error eventName & " - event registry not set on this thread"
+  else:
```
```suggestion
  if regPtr.isNil():
    chronicles.error eventName & " - event registry not set on this thread"
    return
```

ffi/ffi_events.nim
```diff
@@ -149,3 +172,105 @@ proc snapshotListeners*(
...
+                eventType: eventName, payload: eventPayload
+              )
+            )
+            defer:
+              cborFreeShared(data)
+            for listener in snap:
+              listener.callback(
+                RET_OK, cast[ptr cchar](data), cast[csize_t](dataLen), listener.userData
+              )
+          except Exception, CatchableError:
```
gmelodie: Logic here is very similar to that of `onNotResponding` from ffi_context, perhaps we could extract and reuse

gmelodie (approved):

ffi/ffi_context.nim
Ivansete-status: Addressed in 65c54adf974b18b8df35fbf1f34c9e87c3cf827b

ffi/ffi_context.nim
Ivansete-status: Addressed in 65c54adf974b18b8df35fbf1f34c9e87c3cf827b

ffi/ffi_events.nim
Ivansete-status: This is what claude said:
> it doesn't add real safety either: if another thread might still touch reg, it could re-acquire between withLock exit and deinitLock — still UB. So the lock buys nothing here; it's just extra instructions.

ffi/ffi_events.nim
Ivansete-status: ouch! addressed in c254f0923a8a63188714fbb349d4051918de2081

ffi/ffi_events.nim
Ivansete-status: Addressed in 516ad682c7a95fc599c72c2840d9dfc3f2370f1e

ffi/ffi_events.nim
Ivansete-status: Addressed in 516ad682c7a95fc599c72c2840d9dfc3f2370f1e

PR #50, opened 2026-05-26
Event listener abi

Ivansete-status:
Replace the legacy single-slot event ABI with the multi-listener one.

(Stacked on #49.)

- `declareLibrary` macro now emits `_add_event_listener(ctx, event_name, cb, ud) -> uint64` and `_remove_event_listener(ctx, listener_id) -> cint`. `_set_event_callback` is removed end-to-end.
- C++ and Rust codegen keep the existing `setEventHandlers(Events)` / `set_event_handlers(Events)` shape on the foreign side but route through `_add_event_listener(ctx, "", &eventTrampoline, events_)` internally. Re-installation drops the previous listener via `_remove_event_listener` first so the registry never holds a dangling pointer into a freed `Events` box. A new `event_listener_id_` / `event_listener_id` field tracks the active listener id.
- Regenerated `my_timer.hpp` / `api.rs` / `ffi.rs` reflect the new ABI; the existing C++ e2e (16/16) and unit tests stay green.

Issue: #40

### DIFF
--- a/ffi/ffi_events.nim
+++ b/ffi/ffi_events.nim
@@ -59,25 +59,12 @@ const WildcardEventName* = ""
 # ---------------------------------------------------------------------------
 
 proc initEventRegistry*(reg: var FFIEventRegistry) =
-  ## Must be called exactly once on the owning thread before the registry
-  ## is shared. The embedded `Lock` wraps a platform primitive that cannot
-  ## be safely double-initialised, so concurrent callers would hit UB at
-  ## the OS layer — the lock itself can't defend against its own init.
   reg.lock.initLock()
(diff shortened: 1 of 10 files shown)


fcecin (approved):
LGTM

ffi/ffi_events.nim
```diff
@@ -106,6 +106,8 @@ proc addEventListener*(
     if eventName.len == 0:
       reg.wildcard.add(listener)
     else:
+      # `mgetOrPut` lets us avoid a `[]` lookup that the effect tracker
+      # would otherwise see as raising `KeyError`.
```
```suggestion
```

gmelodie (approved):

PR #51, opened 2026-05-26
Cpp typed event listeners

Ivansete-status:
Replace the bundled-handler `setEventHandlers(Events)` C++ API with per-event typed listener methods and a generic wildcard, all on top of the new `_add_event_listener` ABI.

Drops the `Events` struct + `setEventHandlers`. Emits per declared `{.ffiEvent.}`:
- `ListenerHandle addOn<NimProc>Listener(std::function<void(const T&)>)` — typed at compile time; the per-listener trampoline decodes the CBOR envelope's `payload` field as `T`.
- `ListenerHandle addEventListener(std::function<void(int retCode, const std::string& eventId, const std::vector<uint8_t>& envelope)>)` — catch-all wildcard; the trampoline pre-extracts `eventType` from the envelope so consumers branch by event name.
- `bool removeEventListener(ListenerHandle)`.
- New `decodeEventPayload<T>(envelope, out)` helper lifts a wildcard envelope into any registered event payload type.

Listeners are owned by a `std::unordered_map<uint64_t, std::unique_ptr<ListenerBase>>` keyed by id, so the raw pointer handed to the dylib as `user_data` stays valid until `removeEventListener` drops it.

`examples/timer/cpp_bindings/main.cpp` gets a section [7] demoing the new API.

`tests/e2e/cpp/test_timer_e2e.cpp` rewrites the existing event test and adds `MultipleTypedListenersAllFire`, `RemoveEventListenerStopsDelivery`, and `WildcardListenerReceivesEventIdAndDecodesPayload`. 19/19 green.

Note: diff is ~470 lines added (≈360 net), slightly over the 400-line target, because the codegen rewrite, regenerated header, and reworked e2e tests are tightly coupled at the commit level.

Issue: #40

### DIFF
--- a/examples/timer/cpp_bindings/my_timer.hpp
+++ b/examples/timer/cpp_bindings/my_timer.hpp
@@ -12,11 +12,11 @@
 #include <optional>
 #include <type_traits>
 #include <cstring>
-#include <unordered_map>
 extern "C" {
 #include <tinycbor/cbor.h>
 }
 
+#include <unordered_map>
 // ── encode_cbor overloads (primitives + containers) ─────────────────────
 // Per-struct encode_cbor / decode_cbor are emitted by cpp.nim next to each
 // generated struct; these helpers cover the leaf types they defer into.
--- a/examples/timer/cpp_bindings/my_timer.hpp
+++ b/examples/timer/cpp_bindings/my_timer.hpp
@@ -857,10 +857,8 @@ private:
         if (!ud || ret != 0 || !msg || len == 0) return;
         auto* listener = static_cast<TypedListener<T>*>(ud);
         if (!listener->fn) return;
-        std::vector<std::uint8_t> bytes(reinterpret_cast<const std::uint8_t*>(msg),
-                                        reinterpret_cast<const std::uint8_t*>(msg) + len);
         CborParser parser; CborValue it;
-        if (cbor_parser_init(bytes.data(), bytes.size(), 0, &parser, &it) != CborNoError) return;
+        if (cbor_parser_init(reinterpret_cast<const std::uint8_t*>(msg), len, 0, &parser, &it) != CborNoError) return;
         if (!cbor_value_is_map(&it)) return;
         CborValue payloadField;
         if (cbor_value_map_find_value(&it, "payload", &payloadField) != CborNoError) return;
--- a/examples/timer/cpp_bindings/my_timer.hpp
+++ b/examples/timer/cpp_bindings/my_timer.hpp
@@ -1,4 +1,9 @@
 #pragma once
+// Generated bindings require C++20 — the event-listener API uses
+// std::span<const std::uint8_t> for the wildcard callback.
+#if !defined(__cplusplus) || __cplusplus < 202002L
+#error "nim-ffi generated headers require C++20 or later"
+#endif
 #include <string>
 #include <cstdint>
 #include <chrono>
@@ -17,6 +22,7 @@ extern "C" {
 }
 
 #include <unordered_map>
+#include <span>
 // ── encode_cbor overloads (primitives + containers) ─────────────────────
 // Per-struct encode_cbor / decode_cbor are emitted by cpp.nim next to each
 // generated struct; these helpers cover the leaf types they defer into.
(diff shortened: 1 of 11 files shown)


gmelodie (approved):

fcecin (approved):
LGTM

ffi/codegen/cpp.nim
```diff
@@ -138,88 +138,191 @@ proc cppBracedInit(structName: string, fieldNames: seq[string]): string =
...
+  lines.add("    template <class T>")
+  lines.add("    static void typedTrampoline(int ret, const char* msg, std::size_t len, void* ud) {")
+  lines.add("        if (!ud || ret != 0 || !msg || len == 0) return;")
+  lines.add("        auto* listener = static_cast<TypedListener<T>*>(ud);")
+  lines.add("        if (!listener->fn) return;")
+  lines.add(
+    "        std::vector<std::uint8_t> bytes(reinterpret_cast<const std::uint8_t*>(msg),"
+  )
+  lines.add(
+    "                                        reinterpret_cast<const std::uint8_t*>(msg) + len);"
```
NagyZoltanPeter: This seems an unnecessary copy.

ffi/codegen/cpp.nim
```diff
@@ -138,88 +138,191 @@ proc cppBracedInit(structName: string, fieldNames: seq[string]): string =
...
+  )
   lines.add("        CborParser parser; CborValue it;")
-  lines.add("        if (cbor_parser_init(bytes.data(), bytes.size(), 0, &parser, &it) != CborNoError) return;")
+  lines.add(
+    "        if (cbor_parser_init(bytes.data(), bytes.size(), 0, &parser, &it) != CborNoError) return;"
```
NagyZoltanPeter:
```suggestion
    "        if (cbor_parser_init(reinterpret_cast<const std::uint8_t*>(msg), len, 0, &parser, &it) != CborNoError) return;"
```

ffi/codegen/cpp.nim
```diff
@@ -138,88 +138,191 @@ proc cppBracedInit(structName: string, fieldNames: seq[string]): string =
...
+  lines.add("                CborValue evtField;")
+  lines.add(
+    "                if (cbor_value_map_find_value(&it, \"eventType\", &evtField) == CborNoError"
+  )
+  lines.add("                    && cbor_value_is_text_string(&evtField)) {")
+  lines.add("                    (void)decode_cbor(evtField, eventId);")
+  lines.add("                }")
+  lines.add("            }")
+  lines.add("        }")
+  lines.add("        listener->fn(ret, eventId, envelope);")
```
NagyZoltanPeter, line 322: If I understand it correctly this is just a flat dispatcher for further processing. Is it?
As this call `listener->fn(...)` is a sync call I would rather use `std::span<uint_8>` instead of `std::vector<uint_8>`.
Because we already have the allocated buffer pointed by `*msg`, std::span makes it usable in C++ as a view instead of a copy.

examples/timer/cpp_bindings/my_timer.hpp
```diff
@@ -809,38 +837,64 @@ class MyTimerCtx {
...
+                    (void)decode_cbor(evtField, eventId);
+                }
             }
         }
+        listener->fn(ret, eventId, envelope);
     }
 
+    void* ptr_;
+    std::chrono::milliseconds timeout_;
+    std::unordered_map<std::uint64_t, std::unique_ptr<ListenerBase>> listeners_;
```
NagyZoltanPeter, line 908: any map considering few items will be slower than a simple vector scan - even more performant if resized at initialization to a rule of thumb estimated item count.
Of course vector<tuple<std::uint64_t, std::unique_ptr<ListenerBase>> maybe less convenient to use.
So take it as a side comment - little bit of premature optimization.

NagyZoltanPeter (approved):
Some nitpicks added. Otherwise LGTM.

ffi/codegen/cpp.nim
Ivansete-status: Thanks! Addressed in d025ec5c021050fa01e3d443cbd8c0fbdf63415d

ffi/codegen/cpp.nim
Ivansete-status: Addressed in d025ec5c021050fa01e3d443cbd8c0fbdf63415d

ffi/codegen/cpp.nim
Ivansete-status: This "wildcard" event dispatcher is meant to mimic the original `set_event_callback` which allowed to subscribed to all lib events at once.

Thanks for the recommendation. Addressed in fc27e309fcddd759f4e4aa210a0b904814087d38

EDITED: I think is fair to set minimum supported C++20 as it is the most used in logos modules, afaik.

PR #52, opened 2026-05-26
rust codegen: per-event typed add_on_<x>_listener + wildcard add_event_listener

Ivansete-status:
Mirror of the C++ codegen rewrite (#51) for the Rust side. Replaces `set_event_handlers(Events)` with per-event typed listener methods and a generic wildcard, all on top of the `_add_event_listener` / `_remove_event_listener` ABI introduced in #50.

Drops the `Events` struct + `set_event_handlers`. Emits per declared `{.ffiEvent.}`:
- `pub fn add_on_<nim_proc>_listener<F: Fn(&T) + Send + Sync + 'static>(&self, handler: F) -> ListenerHandle` — typed at compile time; the per-listener `extern "C"` trampoline deserialises the CBOR envelope's `payload` field via `ciborium`.
- `pub fn add_event_listener<F: Fn(c_int, &str, &[u8]) + Send + Sync + 'static>(&self, handler: F) -> ListenerHandle` — catch-all wildcard; the trampoline pre-extracts `eventType` so consumers branch by event name.
- `pub fn remove_event_listener(&self, handle: ListenerHandle) -> bool`.

New `decode_event_payload::<T>(envelope)` helper for wildcard handlers that want a typed payload.

Per-listener `*Handler` structs own the closure box; the lib's raw `user_data` pointer stays valid because the box lives in `Mutex<HashMap<u64, Box<dyn Any + Send>>>` on the ctx until `remove_event_listener` drops it.

Regenerated `api.rs` / `ffi.rs` / `Cargo.lock`.

Issue: #40

### DIFF
--- a/examples/timer/rust_bindings/src/api.rs
+++ b/examples/timer/rust_bindings/src/api.rs
@@ -211,6 +211,22 @@ impl MyTimerCtx {
         Ok(Self { ptr: addr as *mut c_void, timeout, listeners: std::sync::Mutex::new(std::collections::HashMap::new()) })
     }
 
+    fn add_listener_inner(
+        &self,
+        event_name: *const c_char,
+        callback: ffi::FFICallback,
+        raw: *mut c_void,
+        owned: Box<dyn std::any::Any + Send>,
+    ) -> ListenerHandle {
+        let id = unsafe {
+            ffi::my_timer_add_event_listener(self.ptr, event_name, callback, raw)
+        };
+        if id != 0 {
+            self.listeners.lock().unwrap().insert(id, owned);
+        }
+        ListenerHandle { id }
+    }
+
     /// Register a typed listener for `on_echo_fired`. The returned handle can be
     /// passed to `remove_event_listener` to unregister.
     pub fn add_on_echo_fired_listener<F>(&self, handler: F) -> ListenerHandle
@@ -218,15 +234,7 @@ impl MyTimerCtx {
     {
         let owned: Box<OnEchoFiredHandler> = Box::new(OnEchoFiredHandler { f: Box::new(handler) });
         let raw = &*owned as *const OnEchoFiredHandler as *mut c_void;
-        let id = unsafe {
-            ffi::my_timer_add_event_listener(
-                self.ptr, b"on_echo_fired\0".as_ptr() as *const c_char,
-                on_echo_fired_trampoline, raw)
-        };
-        if id != 0 {
-            self.listeners.lock().unwrap().insert(id, owned);
-        }
-        ListenerHandle { id }
+        self.add_listener_inner(b"on_echo_fired\0".as_ptr() as *const c_char, on_echo_fired_trampoline, raw, owned)
     }
 
     /// Register a catch-all listener that receives every event.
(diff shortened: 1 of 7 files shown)


fcecin (approved):
LGTM

ffi/codegen/rust.nim
```diff
@@ -446,77 +446,105 @@ proc generateApiRs*(
...
   if events.len > 0:
-    lines.add("/// Typed event handlers for `$1`. Each field is `None` by" % [ctxTypeName])
-    lines.add("/// default; set the ones you care about and pass to")
-    lines.add("/// `$1::set_event_handlers`." % [ctxTypeName])
-    lines.add("#[allow(non_snake_case)]")
-    lines.add("pub struct Events {")
-    lines.add("    pub on_error: Option<Box<dyn Fn(&str) + Send + Sync>>,")
     for ev in events:
+      let handlerStruct = capitalizeFirstLetter(ev.nimProcName) & "Handler"
+      let trampolineName = camelToSnakeCase(ev.nimProcName) & "_trampoline"
```
gmelodie, line 457: We could introduce `heck` as a dependency (it seems to be used by important crates like `serde`) to do `ev.nimProcName.to_snake_case`

gmelodie: Ah! Since we already include serde, that should be the solution of choice i think

examples/timer/rust_bindings/src/api.rs
```diff
@@ -203,30 +208,57 @@ impl MyTimerCtx {
...
+            ffi::my_timer_add_event_listener(
                 self.ptr, b"\0".as_ptr() as *const c_char,
-                my_timer_event_trampoline, raw as *mut c_void);
+                my_timer_wildcard_trampoline, raw)
+        };
+        if id != 0 {
+            self.listeners.lock().unwrap().insert(id, owned);
         }
+        ListenerHandle { id }
+    }
```
gmelodie, line 251: There seems to be some repetition here, maybe we could improve it? I'm reading the generated code as it's more straightforward to review, but these changes should be applied on the generation, not here per se

ffi/codegen/rust.nim
gmelodie: Ps: same for other string manipulation functions. rust has a richer ecosystem than nim, we should use that

ffi/ffi_events.nim
```diff
@@ -106,6 +106,8 @@ proc addEventListener*(
     if eventName.len == 0:
       reg.wildcard.add(listener)
     else:
+      # `mgetOrPut` lets us avoid a `[]` lookup that the effect tracker
+      # would otherwise see as raising `KeyError`.
```
gmelodie: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaargh!
```suggestion
```

gmelodie (approved):

ffi/codegen/rust.nim
Ivansete-status: These case-conversion calls (`camelToSnakeCase`, `capitalizeFirstLetter`) run in the **Nim codegen** at compile time — their output is baked into the generated Rust source as static identifiers (`OnEchoFiredHandler`, `on_echo_fired_trampoline`, etc.). By the time the generated crate sees those names they're already plain `ident` tokens, so a Rust crate like `heck` has nothing to run on. Adding `heck` to the generated crate's `Cargo.toml` wouldn't be reachable from the Nim codegen either.

(And `serde` only does serialization — `heck` is the actual case-conversion crate.)

I'd resolve this thread unless you see a different angle.

( the above reply is auto-generated by Claude )

ffi/ffi_events.nim
Ivansete-status: omg, sorry. That comes from stacked PR

examples/timer/rust_bindings/src/api.rs
Ivansete-status: Simplified in 2b51c6e9071453707b2f0f56f6172e70a6cd11ac

PR #53, opened 2026-05-26
Rust event examples

Ivansete-status:
Ship two bundled examples under the generated Rust crate so foreign-side users see the listener API in action.

**Stacked on the Rust codegen PR (#52).** Base is `rust-typed-event-listeners`; GitHub will retarget down the chain as upstream PRs land.

- `examples/main.rs` — sync flow using `std::sync::mpsc` to bridge a typed `add_on_echo_fired_listener` into `main`, plus a wildcard `add_event_listener` that uses `decode_event_payload::<EchoEvent>(envelope)` for the matching event id.
- `examples/tokio_main.rs` — same shape via `#[tokio::main]` + `tokio::sync::mpsc::unbounded_channel`, bounded `tokio::time::timeout` on the wait.
- `generateCargoToml` now emits a `[dev-dependencies]` block with tokio's `rt-multi-thread` + `macros` features so the bundled examples can use `#[tokio::main]` without polluting the library's runtime profile.

Run with `cargo run --example main` / `cargo run --example tokio_main` (set `DYLD_LIBRARY_PATH=<repo>` on macOS until `build.rs` emits an rpath; tracked as a follow-up).

Issue: #40

### DIFF
--- a/examples/timer/rust_bindings/Cargo.lock
+++ b/examples/timer/rust_bindings/Cargo.lock
@@ -181,6 +181,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8fc7f01b389ac15039e4dc9531aa973a135d7a4135281b12d7c1bc79fd57fffe"
 dependencies = [
  "pin-project-lite",
+ "tokio-macros",
+]
+
+[[package]]
+name = "tokio-macros"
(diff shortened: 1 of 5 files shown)


fcecin (approved):
LGTM

gmelodie (approved):

PR #55, opened 2026-05-31
feat(ffi): C/Go binding generators + native (zero-serialization) C ABI

Ivansete-status:
First of a four-PR stack splitting the native-FFI work into reviewable units.
Each PR targets the previous one; this one targets `master`.

### What
- **C and Go (cgo) binding generators** alongside the existing Rust/C++ ones.
- **Native, zero-serialization C ABI** emitted *next to* the CBOR ABI: every
  `{.ffi.}`/`{.ffiCtor.}` proc is exported twice — `<name>` (native, typed args
  by value, result via callback) and `<name>_cbor` (CBOR request/response) —
  from the *same* library, chosen per call site.
- Test fix: point the C-shape integration tests at the renamed
  `*CborExport`/`*CborCtorExport` entry points the split introduced.

### Stack
1. **this PR** — generators + native C ABI
2. `ffi-native-pod-params` — struct/seq/Option params cross natively (POD)
3. `ffi-go-struct-params` — Go bindings for those params
4. `ffi-native-typed-returns` — native typed struct returns (C + Go) + stress tests

> Note: conceptually related to the draft #29 "introduce raw FFI mode" but a
> separate lineage based on current `master`; worth reconciling during review.


gmelodie:
Replaced by cwire

PR #56, opened 2026-05-31
feat(ffi): cross struct/seq/Option params natively (POD) + C & IPC examples

Ivansete-status:
PR 2/4 of the native-FFI stack. **Base: `ffi-native-c-abi` (#55).**

### What
- Wire the generated **POD machinery** into the native param path: a registered
  `{.ffi.}` struct now crosses the FFI-thread boundary as its `<T>Pod` mirror —
  `clonePod` deep-copies it off the caller's buffers, `podToNim` rebuilds it on
  the FFI thread, `freePod` releases it. Strings → `cstring`; the `<T>Pod`
  `{.bycopy.}` layout is identical to the C-header struct, so the `exportc`
  symbol's ABI matches the header. Applies to `{.ffi.}` and `{.ffiCtor.}`.
- **Native C example** (`examples/timer/c_bindings`): runnable driver + Makefile.
- **IPC example** (`examples/timer/ipc`): CBOR-over-socket server/client showing
  the cross-process / cross-machine path (same binaries do AF_UNIX or TCP).

### Validation
End-to-end from C (TimerConfig, EchoRequest, nested ComplexRequest with
seq-of-structs / seq-of-strings / two Options) and ASAN-clean.


gmelodie:
Replaced by cwire

PR #57, opened 2026-05-31
feat(codegen): Go bindings for struct/seq/Option params + example

Ivansete-status:
PR 3/4 of the native-FFI stack. **Base: `ffi-native-pod-params` (#56).**

### What
- The Go generator no longer `// SKIPPED`s procs with struct/seq/Option params.
  It emits an idiomatic Go struct per `{.ffi.}` type plus a `toC()` that marshals
  it into the matching `C.<Type>` (C.CString for strings, C arrays for seqs,
  present-flags for options, recursively for nested structs) and frees the C-side
  allocations after the call (the native path deep-copies, so that's safe).
- The package is self-contained: the native `<lib>.h` is emitted beside the
  `.go` and the cgo directives use `${SRCDIR}`. `genbindings_go` runs gofmt.
- **Native Go example** exercising Echo/Complex/Schedule with idiomatic Go types.

### Validation
`go run` end-to-end; mapping: `seq[T]`→`[]T`, `Option[T]`→`*T`, nested structs.


gmelodie:
Replaced by cwire

PR #58, opened 2026-05-31
feat(ffi): native typed struct returns (C + Go) + stress tests + docs

Ivansete-status:
PR 4/4 of the native-FFI stack. **Base: `ffi-go-struct-params` (#57).**

### What
- **Native typed struct returns.** A `{.ffi.}` proc returning a registered struct
  now delivers it natively instead of CBOR: the handler builds the return's
  `<T>Pod` on the heap (`nimToPod`); the callback gets a typed `const <T>*`
  (len = sizeof); `handleRes` deep-frees it right after the callback —
  callback-lifetime ownership, the caller frees nothing.
- **C**: example casts the callback `msg` to `const <Type>*` and reads it.
- **Go**: struct-returning methods now return typed Go structs. Since the POD is
  freed right after the callback, the POD→Go conversion (`fromC`) runs in an
  exported Go result callback, delivered to the waiting goroutine via a
  `runtime/cgo.Handle` boxed in a C allocation (checkptr/-race safe).
- **Docs**: spell out that the CBOR ABI is for **inter-process comms only** —
  native is the default for same-process callers.
- **Stress harness** (`tests/e2e/c`): many threads hammer one context,
  alternating native (typed struct in/out) and CBOR.

### Validation
C typed returns ASAN-clean; Go typed returns clean under `go run -race`; stress
harness clean at 6 threads × 1500 iters (9000 calls/ABI) under **ASAN and TSAN**
— no leaks, use-after-free, or data races.


gmelodie:
Replaced by cwire

PR #59, opened 2026-05-31
feat(codegen): native Swift generator + iOS example over the C ABI

Ivansete-status:
Adds an **iOS example** consuming the timer library's native (zero-serialization)
C ABI from Swift. Based on the native-FFI stack (#58) so it can use typed struct
returns.

### What
- `examples/timer/ios/` — a SwiftPM package with an idiomatic `TimerNode` Swift
  wrapper over the native ABI. Struct returns (`EchoResponse`) come back as typed
  Swift values, read out of the C-POD inside the result callback.
- `build-xcframework.sh` cross-compiles the Nim library to a static
  `MyTimer.xcframework` with **ios-arm64** (device), **ios-arm64-simulator**, and
  **macos-arm64** slices. The xcframework is assembled by hand (no
  `xcodebuild -create-xcframework`), so it builds headless / in CI.

### Validation
- Cross-compiles cleanly for device + simulator (correct arm64 slices, native +
  CBOR symbols present).
- The macos-arm64 slice makes the wrapper host-testable: **`swift test` passes**
  (create / version / typed `echo` / a 200-iteration loop).

> Separate PR, as requested. Sibling to the Android example PR; both target the
> stack tip and retarget to `master` once it lands.


Ivansete-status:
Expanded scope: this PR now also adds the **Swift generator** that produces the wrapper, not just the hand-written example.

- `ffi/codegen/swift.nim` — emits an idiomatic Swift wrapper over the native C ABI (imports the `CMyTimer` clang module, never touches CBOR). Wired into the `targetLang` dispatch + a `genbindings_swift` nimble task.
- `Sources/MyTimer/MyTimer.swift` is now **generated** (was hand-written); tests retargeted to the derived `MyTimerNode`.
- Callback-shape selection (ack / string / struct) is chosen per proc from `FFIProcMeta`; struct returns are copied out inside the callback (deep-free-after-callback rule).
- Procs needing `seq`/`Option` or multi-struct param marshaling (`complex`, `schedule`) are skipped with a logged notice so the wrapper always compiles — that marshaling, events, and async are the next increments.

Validated: `build-xcframework.sh` + `swift test` pass 2/2 on the macOS slice.

PR #60, opened 2026-05-31
feat(codegen): native Kotlin/JNI generator + Android example over the C ABI

Ivansete-status:
Adds an **Android example** consuming the timer library's native (zero-
serialization) C ABI from Kotlin via JNI. Based on the native-FFI stack (#58) so
it can use typed struct returns.

### What
- `examples/timer/android/` — an Android library module with an idiomatic Kotlin
  `TimerNode` class.
- `jni/my_timer_jni.c` — a JNI shim that turns each Kotlin `external fun` into a
  blocking call into the native ABI and reads the typed `EchoResponse` struct out
  of the result callback.
- `build-libs.sh` cross-compiles, per ABI (**arm64-v8a** + **x86_64**), both
  `libmy_timer.so` (the Nim library) and `libmy_timer_jni.so` (the bridge) into
  `src/main/jniLibs/`, which Gradle packages automatically. An instrumented test
  covers create / version / echo.

### Validation
Native build verified for both ABIs: correct aarch64/x86_64 ELF, the 4
`Java_org_logos_mytimer_*` symbols exported, and `libmy_timer_jni.so` correctly
`NEEDS` `libmy_timer.so`. (Kotlin/Gradle layer runs on a device/emulator.)

> Separate PR, as requested. Sibling to the iOS example PR; both target the stack
> tip and retarget to `master` once it lands.


Ivansete-status:
Expanded scope: this PR now also adds the **Kotlin/JNI generator** that produces the wrapper, not just the hand-written example (mirrors what #59 did for Swift).

- `ffi/codegen/kotlin.nim` — from one set of `{.ffi.}` metadata it emits **both** artifacts: the idiomatic Kotlin `AutoCloseable` wrapper (`MyTimerNode.kt`) and the JNI shim (`my_timer_jni.c`). Wired into the `targetLang` dispatch + a `genbindings_kotlin` nimble task.
- The hand-written `TimerNode.kt` / `my_timer_jni.c` are replaced by generated output; the instrumented test points at the derived `MyTimerNode`.
- Callback-shape selection (ack / string / struct) is per proc; the shim blocks on a condvar until the FFI-thread callback fires. An all-string struct return crosses as a `String[]` repacked into a `data class`.
- Procs needing `seq`/`Option` params, multiple struct params, or a non-string struct return (`complex`, `schedule`) are skipped with a logged notice so the generated code always compiles — that marshaling, events, and async are the next increments.

Validated by cross-compiling both ABIs with the NDK: **arm64-v8a + x86_64 build clean**, the four `Java_org_logos_mytimer_MyTimerNode_*` symbols are exported, and `libmy_timer_jni.so` correctly `NEEDS` `libmy_timer.so`.

PR #61, opened 2026-05-31
docs(examples): native (same-process) C++ example

Ivansete-status:
Adds the **native C++ example** to sit alongside the existing CBOR C++ bindings,
so C++ has both ABIs covered (as requested). Based on the native-FFI stack (#58)
for typed struct returns.

### What
- `examples/timer/cpp_native/` — a C++ program that links the library and calls
  the **native** `<name>` ABI directly: `{.ffi.}` structs by value in, typed
  `EchoResponse` struct out (read in the callback). No CBOR, no tinycbor. An RAII
  `TimerNode` wrapper bridges the async FFI-thread callback to a synchronous API
  with `std::future`.

### Native vs CBOR C++ (now both present)
| Example | ABI | For |
|---------|-----|-----|
| `cpp_native/` (new) | native `<name>` | same-process / local |
| `cpp_bindings/` | CBOR `<name>_cbor` | inter-process |

### Validation
Builds and runs end-to-end (`make run`): version + typed echo round-trip.

> Note: PR #55 carries a companion fix pointing the **CBOR** C++ generator at the
> `<name>_cbor` symbols (it was calling the native symbol with CBOR args after
> the ABI split — that had reddened the C++ e2e). With both, the CBOR and native
> C++ paths work side by side.

### DIFF
--- a/examples/timer/cpp_native/.gitignore
+++ b/examples/timer/cpp_native/.gitignore
@@ -1,0 +1,3 @@
+/example
+/libmy_timer.dylib
+/libmy_timer.so
--- a/examples/timer/cpp_native/Makefile
+++ b/examples/timer/cpp_native/Makefile
@@ -1,0 +1,42 @@
+# Build the native (same-process) C++ example for the timer library.
+#
+#   make run     # build the Nim dylib + the C++ driver, then run it
+#   make clean
+#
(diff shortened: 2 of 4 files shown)


Ivansete-status:
Closing as this it is already covered by https://github.com/logos-messaging/nim-ffi/pull/63

PR #62, opened 2026-05-31
feat(ffi): native event delivery + dual-ABI event symbols (C + Go)

Ivansete-status:
## What

Foundation for emitting both the **native** (zero-serialization C-POD) and **CBOR** ABIs side-by-side for `{.ffiEvent.}` events, completing the request-side dual ABI from the earlier PRs in this stack.

- **Dual event dispatch** — one `{.ffiEvent.}` declaration now fires both a native typed-POD listener and a CBOR listener from a single dispatch. Native listeners get the payload built once via `nimToPod`/`freePod`; CBOR is serialized once for the rest.
- **Dual-ABI symbol naming** mirrors requests: bare `<lib>_add_event_listener` is native, `<lib>_add_event_listener_cbor` is CBOR.
- **`-d:ffiMode=native|cbor|both`** strdefine (default `both`) gates which ABI each generator emits, with `genbindings_*_native` / `_cbor` nimble tasks.
- **C** generator documents native event payloads (`"on_echo_fired" -> const EchoEvent *`) and routes the CBOR header to `_add_event_listener_cbor`.
- **Go** generator emits per-event typed native handlers `On<Event>(func(<Payload>))` with exported callbacks + `fromC`.

Native is the same-process path; CBOR is IPC-only.

## Test

`test_native_events`, `test_native_pod_types`, and the C native+CBOR concurrency stress harness pass (ASAN/TSAN clean).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #63, opened 2026-05-31
feat(codegen): native C++ generator (typed requests + events)

Ivansete-status:
## What

Native (zero-serialization) C++ generator, emitted alongside the CBOR C++ bindings — the C++ counterpart of the native C/Go work in this stack.

- Emits `<lib>_native.hpp` (plus the native `<lib>.h` via the C generator) with idiomatic C++ types, `toC`/`fromC`, and a `<Lib>Node` that blocks on `std::future`.
- **Holder pattern** for native marshalling: a holder owns the backing storage (`std::vector` arrays, strings borrow the source via `.c_str()`) while the C-POD struct's pointers reference it for the call duration (the lib deep-copies).
- **Native typed events** — `On<Event>(std::function<void(const Payload&)>)` registered through the bare native listener entry point, with `extern "C"` trampolines reading the raw C-POD and a `std::map<uint64_t, unique_ptr<ListenerBase>>` owning the handlers.
- Static C++/CMake blocks live in `ffi/codegen/templates/cpp/*.tpl` (via `staticRead`).

Builds on #62 (native event delivery).

## Test

C++ e2e suite passes; the holder memory pattern is ASAN-clean.


Ivansete-status:
Closing as master already has cpp.nim generator.

PR #64, opened 2026-05-31
feat(codegen): native Rust generator (typed requests + events)

Ivansete-status:
## What

Native (zero-serialization) Rust generator, emitted alongside the CBOR Rust crate — the Rust counterpart of the native C/Go/C++ work in this stack. `rust.nim` is split into `rust_cbor.nim` + `rust_native.nim` (mirroring `cpp.nim`/`cpp_native.nim`); the native and CBOR crates write to separate output dirs (`rust_native_bindings` vs `rust_bindings`).

- **`<lib>_native` crate** (`ffi.rs` / `types.rs` / `api.rs` / `lib.rs`) with `#[repr(C)]` POD mirrors, idiomatic structs, and `to_c`/`from_c`.
- **Requests**: scalars, strings, bools, floats, nested structs, **sequences (`Vec`) and optionals (`Option`)**. `to_c` returns a holder owning the `CString`s and C-array backing on the heap, so the C struct's raw pointers survive the move and stay valid for the call (the lib deep-copies). `from_c` reads seq via `slice::from_raw_parts` and Option via a present-flag.
- **Native typed events**: per-event `add_<wire>_listener(closure)` registers through the bare native entry point; the `extern "C"` trampoline reads the raw C-POD payload and hands the consumer a borrowed idiomatic value via `from_c` — no CBOR on the hot path. The node owns the boxed closures in a `Mutex<HashMap<id, Box<dyn Any>>>`; `remove_event_listener` drops them.
- **`-d:ffiMode`** gating + `genbindings_rust_native` nimble task. Also includes the `_cbor` request-export fix for the CBOR crate.

Builds on #62 (native event delivery).

## Test

The generated demo round-trips `version` / `echo` / `complex` (seq-of-structs + seq-of-strings + two optionals) / `schedule` (three struct params), fires `on_echo_fired` inline delivering a typed `EchoEvent` with zero CBOR, and removes the listener cleanly — validated against the live dylib.


Ivansete-status:
Closing as it already exists a rust.nim generator in master

PR #65, opened 2026-05-31
refactor(codegen): C++ header names follow native=bare / cbor=_cbor

Ivansete-status:
## What

Reconciles the **C++** header filenames with the C generator and the symbol naming. Native (zero-serialization, same-process) is the bare `<lib>.hpp`; CBOR (inter-process) carries the `_cbor` suffix — matching the C headers (`<lib>.h` / `<lib>_cbor.h`) and the `<name>` / `<name>_cbor` exports.

Previously native was `<lib>_native.hpp` and CBOR was the bare `<lib>.hpp` — backwards from the convention, and a collision risk on the native `<lib>.h` when both ABIs emit into one dir. With the flip, a single `genbindings_cpp` run drops `<lib>.hpp` (native) + `<lib>_cbor.hpp` (CBOR) side by side, exactly like `c_bindings` holds both `.h` headers.

Consumers updated: the CBOR cpp example driver and the C++ e2e suite include `*_cbor.hpp`; the native example includes the bare `<lib>.hpp`.

## Test

- Native C++ example builds/runs on `my_timer.hpp`.
- C++ e2e suite: **19/19** on the `_cbor.hpp` headers.
- `check_bindings_cpp` regen is deterministic.

Stacked on #63.


Ivansete-status:
Closing as the suggested changes are outdated. @gmelodie - reopen if consider necessary.

PR #66, opened 2026-05-31
refactor(codegen): Rust crate names follow native=bare / cbor=_cbor

Ivansete-status:
## What

Reconciles the **Rust** crate names with the C generator and the symbol naming. The native (zero-serialization, same-process) crate is the bare `<lib>`; the CBOR (inter-process) crate carries the `_cbor` suffix — mirroring `<lib>.h` / `<lib>_cbor.h` and the `<name>` / `<name>_cbor` exports.

Previously the native crate was `<lib>_native` and the CBOR crate was the bare `<lib>` — backwards from the convention. Only the Cargo package name changes; the linked dylib stays `lib<lib>.dylib` (the `#[link]` name and build.rs are untouched).

Consumers updated: `rust_client` depends on `my_timer_cbor` and imports from it; the native demo imports from the bare `my_timer`.

## Test

- `rust_client` builds against the renamed CBOR crate.
- Native demo round-trips version / echo / event / complex / schedule against the bare crate.
- `check_bindings_rust` regen is deterministic.

Stacked on #64.

### DIFF
--- a/examples/timer/rust_bindings/Cargo.toml
+++ b/examples/timer/rust_bindings/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "my_timer"
+name = "my_timer_cbor"
 version = "0.1.0"
 edition = "2021"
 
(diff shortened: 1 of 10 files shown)


Ivansete-status:
Closing: depends on #64, which is closed. A single Rust wire makes the bare my_timer crate name correct already.

PR #69, opened 2026-06-01
feat: move user event code to a dedicated event thread

### LINKED ISSUE #6: Move user event code to a separate thread

gmelodie:
Closes #6.

## Summary

Move FFI event listener invocation off the FFI thread onto a dedicated event thread fed by a bounded SPSC queue, and fold the standalone watchdog into the same thread via an in-band heartbeat. The wire format and foreign ABI (`<lib>_add_event_listener` / `<lib>_remove_event_listener`) are unchanged — only internals shift.

## What changes

- **`ffi/ffi_events.nim`** — adds a bounded `EventQueue` (POD ring of `c_malloc`-backed payloads, capacity 1024). Dispatch templates now encode and enqueue, never touching `reg.lock`. On queue-full they set the sticky `eventQueueStuck` flag and wake the event thread, which fires `onNotResponding` from its own loop (firing it from the FFI thread would deadlock against a listener back-pressuring under `reg.lock`).
- **`ffi/ffi_context.nim`** — `watchdogThread` → `eventThread`. `eventThreadBody` drains the queue (taking `reg.lock` for the listener fan-out, preserving the UAF contract from PR #39) and checks an `Atomic[int64]` heartbeat the FFI thread advances each iteration. A frozen heartbeat past the 10 s start-grace fires `onNotResponding` (latched). `sendRequestToFFIThread` short-circuits once the stuck flag is set.
- **`ffi/ffi_context_pool.nim`** — `destroyFFIContext` now fully mirrors `initContextResources` (lock + signal fds + queue + registry), so a reused slot doesn't re-init a live lock or leak fds.
- **Tests** — `tests/unit/test_event_thread.nim` (new) covers async delivery on a separate OS thread, slow-listener isolation from FFI round-trip, heartbeat staleness, and queue overflow. `tests/unit/test_event_dispatch.nim` is updated to drive a real `FFIContext` for the lock-during-invocation regression.

## Lock ordering

- Producers (FFI thread, dispatch templates): only `eventQueue.lock`. Never touch `reg.lock`.
- Consumer (event thread): `eventQueue.lock` to dequeue (released immediately), then `reg.lock` for snapshot + invocation. The two are never held simultaneously.
- Foreign threads (`add_event_listener` / `remove_event_listener`): only `reg.lock`. They block while the event thread is mid-invoke, preserving the UAF-close contract.

## Test plan

- [x] `nimble test` passes under both `--mm:orc` and `--mm:refc`.
- [x] `nimble test_cpp_e2e` passes; `nimble genbindings_cpp` produces no diff (no ABI change).
- [x] `nimble test_sanitized` with `NIM_FFI_SAN=tsan NIM_FFI_MM=orc` is race-free.


ffi/ffi_context.nim
```diff
@@ -6,119 +6,85 @@
...
-    # main library object (e.g., Waku, LibP2P, SDS,  the one to be exposed as a library)
+  myLib*: ptr T # main library object (Waku, LibP2P, SDS, …)
   ffiThread: Thread[(ptr FFIContext[T])]
-    # represents the main FFI thread in charge of attending API consumer actions
   eventThread: Thread[(ptr FFIContext[T])]
-    # drains the event queue and runs the FFI-thread heartbeat check
   lock: Lock
   reqChannel: ChannelSPSCSingle[ptr FFIThreadRequest]
-  reqSignal: ThreadSignalPtr # to notify the FFI Thread that a new request is sent
+  reqSignal: ThreadSignalPtr
```
Ivansete-status, line 21: I like all the style and format changes proposals but I think is interesting to have a separate PR so that we can have better focus. Something to double-check once we merge this branch.

ffi/ffi_context.nim
```diff
@@ -6,119 +6,85 @@
...
 const
-  EventThreadTickInterval* = 1.seconds # bounds idle heartbeat check latency
+  EventThreadTickInterval* = 1.seconds
   FFIHeartbeatStartDelay* = 10.seconds # grace window for library startup
   FFIHeartbeatStaleThreshold* = 1.seconds
 
 include ./event_thread
 include ./ffi_thread
 
+template closeAndNil(field: untyped) =
```
Ivansete-status, line 48: Great catch!

ffi/ffi_context_pool.nim
```diff
@@ -19,7 +15,7 @@ proc acquireSlot[T](pool: var FFIContextPool[T]): Result[ptr FFIContext[T], stri
     var expected = false
     if pool.inUse[i].compareExchange(expected, true):
       return ok(pool.slots[i].addr)
-  return err("FFI context pool exhausted (max " & $MaxFFIContexts & " contexts)")
+  err("FFI context pool exhausted (max " & $MaxFFIContexts & " contexts)")
```
Ivansete-status, line 18: nit, I like to be always explicit even though Nim allows that. May apply elsewhere.

ffi/event_thread.nim
```diff
@@ -111,14 +108,19 @@ proc check[T](hb: var HeartbeatMonitor, ctx: ptr FFIContext[T]) =
...
-    # until then the wait always times out and we fall through to the heartbeat check.
+    # Wake on enqueue or tick — whichever first.
     discard await ctx.eventQueueSignal.wait().withTimeout(EventThreadTickInterval)
 
     ctx.drainEventQueue()
 
+    # Fire after drain so reg.lock is free — FFI-thread would deadlock here.
+    if not notifiedStuck and ctx.eventQueueStuck.load():
+      onNotResponding(ctx)
+      notifiedStuck = true
```
Ivansete-status, line 119: Should we also control when `notifiedStuck = false` back again?

ffi/ffi_context.nim
```diff
@@ -137,23 +104,12 @@ proc initContextResources*[T](ctx: ptr FFIContext[T]): Result[void, string] =
...
-
-  ctx.threadExitSignal = ThreadSignalPtr.new().valueOr:
-    return err("couldn't create threadExitSignal ThreadSignalPtr: " & $error)
-
-  ctx.eventQueueSignal = ThreadSignalPtr.new().valueOr:
-    return err("couldn't create eventQueueSignal ThreadSignalPtr: " & $error)
-
-  ctx.eventThreadExitSignal = ThreadSignalPtr.new().valueOr:
-    return err("couldn't create eventThreadExitSignal ThreadSignalPtr: " & $error)
+  newSignalOrErr(ctx.reqSignal, "reqSignal")
```
Ivansete-status, line 107: 👏

ffi/ffi_context.nim
```diff
@@ -180,63 +135,51 @@ proc initContextResources*[T](ctx: ptr FFIContext[T]): Result[void, string] =
   success = true
   ok()
 
+proc fireOrErr(sig: ThreadSignalPtr, name: string): Result[void, string] =
+  let fired = sig.fireSync().valueOr:
+    return err("error signaling " & name & ": " & $error)
```
Ivansete-status: Even though it breaks the sentence I think is interesting having a `:` to separate a continuous hard-coded string from a variable one, considering the main purpose is to ease the future lookups in code. May apply elsewhere.
```suggestion
    return err("error signaling: " & name & ": " & $error)
```

Ivansete-status (approved):
LGTM! Thanks for it! 🙌 
The only interesting suggestion is to perform the format changes in a separate PR so that we have a clear separation of concerns on each.

PR #70, opened 2026-06-02
Remove wildcard event listener; keep per-event dispatch

Ivansete-status:
## What

Completely removes the **wildcard** event-listener concept. Library-initiated events now follow a single model: callers subscribe to each event separately via the per-event typed listeners (`add_on_<x>_listener` / `addOn<X>Listener`). A dispatched event reaches exactly the listeners subscribed to its name.

This drops the catch-all listener (`add_event_listener` with an empty event name, the C++ `addEventListener` overload, and the Rust `add_event_listener`) that received every event as a raw CBOR envelope plus the wire `eventId`.

## Changes

**Nim core**
- `ffi/ffi_events.nim` — removed the `wildcard` registry field and the `WildcardEventName` constant; `addEventListener` always registers under the event name; `removeEventListener` / `removeAllEventListeners` / `snapshotListeners` and the dispatch template no longer fold in wildcard listeners.
- `ffi/ffi_context.nim` — `onNotResponding` no longer concatenates wildcard listeners.
- `ffi/internal/ffi_library.nim` — updated the `_add_event_listener` ABI doc. The C entry point stays: it is the per-event registration primitive the typed listeners call.

**Codegen** (source of truth for generated bindings)
- `ffi/codegen/cpp.nim` — removed the `addEventListener` wildcard method, the `WildcardListener` struct, `wildcardTrampoline`, the `decodeEventPayload<T>` helper, and the now-unused `<span>` include.
- `ffi/codegen/rust.nim` — removed the `WildcardHandler` struct, the wildcard trampoline, `decode_event_payload`, and the `add_event_listener` method.
- C++ templates: comments adjusted (C++20 baseline kept).

**Generated bindings** — regenerated `my_timer.hpp`, `echo.hpp`, and `rust_bindings/` so the checked-in output stays in sync with the codegen.

**Tests & example** — removed wildcard cases from the Nim unit tests (now subscribing to concrete event names), deleted the wildcard C++ e2e test, and rewrote the `main.cpp` demo to use only the typed listener.

## Verification

- Nim unit tests pass under `--mm:orc` and `--mm:refc`.
- C++ e2e suite: 18/18 pass; Rust crate and the C++ example compile; example runs and the typed `onEchoFired` event fires.
- Repo-wide grep confirms zero remaining wildcard references.


ffi/ffi_events.nim
```diff
@@ -120,54 +112,44 @@ proc removeEventListener*(reg: var FFIEventRegistry, id: uint64): bool {.raises:
...
-          listeners.delete(idx)
-          removed = true
-          if listeners.len == 0:
-            emptyKey = key
-            prune = true
-          break
-      if prune:
-        reg.byEvent.del(emptyKey)
+    if prune:
+      reg.byEvent.del(emptyKey)
```
gmelodie: Maybe we could do sth along these lines instead
```nim
withLock reg.lock:
  var
    pruneKey = ""
    prune = false
  for key, listeners in reg.byEvent.mpairs:
    let before = listeners.len
    listeners.keepItIf(it.id != id)
    if listeners.len < before:
      removed = true
      if listeners.len == 0:
        pruneKey = key
        prune = true
      break
  if prune:
    reg.byEvent.del(pruneKey)
```

gmelodie (approved):
`-500` lines!!!

Image

ffi/ffi_events.nim
Ivansete-status: Done in b277ebd — switched to `keepItIf` with a length-delta check, exactly as suggested. Tests green on orc + refc.

Ivansete-status: Addressed in b277ebde751dde170cf28e96676c2fe492d2c46f

PR #71, opened 2026-06-02
refactor(ffi): event thread scaffolding + FFIContext lifecycle split

gmelodie:
Base PR for #6 — splits #69's event-thread work into a context-lifecycle foundation that lands first.

## Summary

Stands up the dedicated event thread and the bounded SPSC `EventQueue` *data structure* alongside the existing dispatch path, without rewiring dispatch onto them yet. The library stays fully functional: the new thread spins up, takes over the heartbeat check the standalone watchdog used to do, drains the (always-empty in this PR) queue each tick, and exits cleanly on stop.

PR #69 builds on this scaffolding and is what actually moves event callbacks off the FFI thread by having `dispatchFFIEvent*` enqueue instead of invoking listeners inline.

## What changes

- `ffi_context.nim` — slimmed down to the `FFIContext` type plus the lifecycle surface (`initContextResources` / `deinitContextResources` / `cleanUpResources` / `signalStop` / `stopAndJoinThreads` / `clearContext`). Now `include`s two new sibling files:
  - `ffi_thread.nim` — `ffiThreadBody`, `sendRequestToFFIThread`, `processRequest`, and the `onFFIThread` threadvar. The FFI thread now advances `ctx.ffiHeartbeat` each iteration.
  - `event_thread.nim` — `eventThreadBody`, the heartbeat monitor, and the liveness events. Emits a typed CBOR `NotRespondingEvent` when the FFI heartbeat goes stale, and a companion `RespondingEvent` when it starts advancing again — both latch so each transition fires at most once per stall episode. Replaces the legacy JSON `onNotResponding` shim.
- `ffi_context_pool.nim` — `destroyFFIContext` mirrors `initContextResources` via the new `deinitContextResources` helper so reused slots don't re-init live locks or leak signal fds.
- `ffi_events.nim` — adds the bounded `EventQueue` type + init/deinit + `tryEnqueueEvent`/`tryDequeueEvent` (POD ring, capacity 1024) so PR #69 has a queue to feed. Adds a shared `notifyListeners(retCode, data, dataLen)` that delivers a non-nil `emptyListenerPayload` sentinel for empty payloads — `std::string(data, len)` / `memcpy` on a nil pointer is UB even at len 0. `notifyListenersErr` becomes a thin adapter. Dispatch templates still call listeners inline (unchanged), now via `notifyListeners`.

## What's intentionally NOT here (lands in #69)

- The dispatch-template rewrite to enqueue instead of invoke inline.
- The `eventQueueStuck` overflow flag and the producer-side back-pressure path (incl. the `sendRequestToFFIThread` short-circuit and the event-thread `onNotResponding` fire on stuck).
- The thread-local hooks bridging dispatch to the queue/signal (`ffiCurrentEventQueue`, `ffiCurrentNotifyEventEnqueued`, etc.).
- Any priority/`EventQueueAlmostFull` semantics — those require the enqueue path that #69 introduces.

## Test plan

- [x] `nimble test` under `--mm:orc` and `--mm:refc`
- [x] `nimble test_cpp_e2e` (no binding diff)
- [x] `nimble test_sanitized` with `NIM_FFI_SAN=tsan NIM_FFI_MM=orc`


ffi/codegen/cpp.nim
```diff
@@ -56,7 +54,7 @@ proc stripLibPrefixCpp(procName, libName: string): string =
   let prefix = libName & "_"
   if procName.startsWith(prefix):
     return procName[prefix.len .. ^1]
-  return procName
+  procName
```
Ivansete-status: nit, I tend to prefer being more explicit and always use `return` term

ffi/ffi_context.nim
```diff
@@ -39,13 +45,62 @@ var onFFIThread* {.threadvar.}: bool
...
+type NotRespondingEvent* = object
+
+const NotRespondingEventName* = "not_responding"
+
+proc encodeNotRespondingEvent(): seq[byte] =
+  EventEnvelope[NotRespondingEvent](
+    eventType: NotRespondingEventName, payload: NotRespondingEvent()
+  ).cborEncode()
+
+proc dispatchToListeners[T](
```
Ivansete-status: This proc is very cool

ffi/ffi_context.nim
```diff
@@ -39,13 +45,62 @@ var onFFIThread* {.threadvar.}: bool
...
+    foreignThreadGc:
+      try:
+        notifyListenersOk(listeners, data, dataLen)
+      except Exception, CatchableError:
+        notifyListenersErr(
+          listeners,
+          "Exception dispatching " & eventName & ": " & getCurrentExceptionMsg(),
+        )
+
+proc onNotResponding*(ctx: ptr FFIContext) =
```
Ivansete-status: Aside from PR's purpose but we may need to also notify when the ffi-thread switches from NotResponding to Responding. Something to consider for future PR.

ffi/ffi_context.nim
```diff
@@ -203,17 +182,29 @@ proc processRequest[T](
...
+    if res.isErr():
+      error "failed to fire eventQueueSignal after enqueue", err = res.error
+
 proc ffiThreadBody[T](ctx: ptr FFIContext[T]) {.thread.} =
   ## FFI thread body that attends library user API requests
   ffiCurrentEventRegistry = addr ctx[].eventRegistry
+  ffiCurrentEventQueue = addr ctx[].eventQueue
+  ffiCurrentEventQueueStuck = addr ctx[].eventQueueStuck
+  ffiEventQueueSignalPtr = ctx.eventQueueSignal
+  ffiCurrentNotifyEventEnqueued = ffiNotifyEventEnqueuedHook
```
Ivansete-status: It seems `ffiCurrentNotifyEventEnqueued` is never invoked.
Similarly for `ffiCurrentEventRegistry`.

ffi/ffi_events.nim
```diff
@@ -142,29 +122,119 @@ proc snapshotListeners*(
...
+    listener.callback(
+      RET_ERR, dataPtr, cast[csize_t](msg.len), listener.userData
+    )
+
+var ffiCurrentEventRegistry* {.threadvar.}: ptr FFIEventRegistry
+var ffiCurrentEventQueue* {.threadvar.}: ptr EventQueue
+var ffiCurrentEventQueueStuck* {.threadvar.}: ptr Atomic[bool]
+var ffiCurrentNotifyEventEnqueued* {.threadvar.}: proc() {.gcsafe, raises: [].}
+  ## Hook (not a queue field) so this module doesn't depend on chronos's
+  ## ThreadSignalPtr. Nil-safe.
```
Ivansete-status: What possible issue are we mitigating?

ffi/ffi_context.nim
```diff
@@ -13,21 +18,22 @@ type FFIContext*[T] = object
     # main library object (e.g., Waku, LibP2P, SDS,  the one to be exposed as a library)
   ffiThread: Thread[(ptr FFIContext[T])]
     # represents the main FFI thread in charge of attending API consumer actions
-  watchdogThread: Thread[(ptr FFIContext[T])]
-    # monitors the FFI thread and notifies the FFI API consumer if it hangs
+  eventThread: Thread[(ptr FFIContext[T])]
+    # drains the event queue and runs the FFI-thread heartbeat check
```
Ivansete-status, line 28: It should also actually handle lib's events. The `dispatchFFIEvent..` should just enqueue the event, from ffi-thread, and make it to be handled by event-thread. The PR needs minimum refactor to achieve that.

ffi/ffi_context.nim
```diff
@@ -13,21 +18,22 @@ type FFIContext*[T] = object
...
     # to signal main thread, interfacing with the FFI thread, that FFI thread received the request
   stopSignal: ThreadSignalPtr
-    # fired by destroyFFIContext so both ffiThread and watchdogThread can exit promptly
-  threadExitSignal: ThreadSignalPtr
-    # fired by ffiThread just before it exits; destroyFFIContext waits on
-    # this with a bounded timeout instead of joining unconditionally, so a
-    # blocked event loop cannot hang the caller forever
+  threadExitSignal: ThreadSignalPtr # bounds destroyFFIContext's wait so a blocked loop cannot hang the caller
+  eventQueueSignal: ThreadSignalPtr # wakes the event thread on enqueue
+  eventThreadExitSignal: ThreadSignalPtr # mirrors threadExitSignal for the event thread
```
Ivansete-status, line 37: FFIContext is getting a little bit too convoluted and difficult to follow. We can split  ffi_context.nim and have ffi_thread.nim and event_thread.nim in future PR.

ffi/ffi_context.nim
```diff
@@ -268,17 +262,99 @@ proc ffiThreadBody[T](ctx: ptr FFIContext[T]) {.thread.} =
...
-        error "draining pending FFI requests on shutdown raised",
-          error = exc.msg
+        error "draining pending FFI requests on shutdown raised", error = exc.msg
 
   waitFor ffiRun(ctx)
 
-proc cleanUpResources[T](ctx: ptr FFIContext[T]): Result[void, string] =
-  ## Full cleanup for heap-allocated contexts: closes all resources and frees memory.
+proc freeQueuedEventPayload(qe: QueuedEvent) =
+  if not qe.name.isNil:
```
Ivansete-status: nit,
```suggestion
  if not qe.name.isNil():
```

ffi/ffi_context.nim
```diff
@@ -268,17 +262,99 @@ proc ffiThreadBody[T](ctx: ptr FFIContext[T]) {.thread.} =
...
+    c_free(cast[pointer](qe.name))
+  if not qe.data.isNil:
+    c_free(qe.data)
+
+proc dispatchQueuedEvent[T](ctx: ptr FFIContext[T], qe: QueuedEvent) =
```
Ivansete-status: Maybe this proc is not needed and just have this logic below.

ffi/ffi_context.nim
```diff
@@ -268,17 +262,99 @@ proc ffiThreadBody[T](ctx: ptr FFIContext[T]) {.thread.} =
...
+  ## Frees `qe`'s c_malloc buffers on exit.
   defer:
-    freeShared(ctx)
+    freeQueuedEventPayload(qe)
+  ctx.dispatchToListeners($qe.name, qe.data, qe.dataLen)
+
+proc drainEventQueue[T](ctx: ptr FFIContext[T]) =
+  while true:
+    let opt = ctx.eventQueue.tryDequeueEvent()
+    if opt.isNone:
```
Ivansete-status: nit, verbs with parenthesis
```suggestion
    if opt.isNone():
```

ffi/ffi_context.nim
```diff
@@ -268,17 +262,99 @@ proc ffiThreadBody[T](ctx: ptr FFIContext[T]) {.thread.} =
...
+      break
+    ctx.dispatchQueuedEvent(opt.get())
+
+type HeartbeatMonitor = object
+  startedAt: Moment
+  lastChange: Moment
+  lastValue: int64
+  notifiedStale: bool
+
+proc initHeartbeatMonitor[T](ctx: ptr FFIContext[T]): HeartbeatMonitor =
```
Ivansete-status: Maybe more idiomatic?
```suggestion
proc init(T: typedesc[HeartbeatMonitor](ctx: ptr FFIContext[TT]): HeartbeatMonitor =
```

ffi/ffi_events.nim
```diff
@@ -142,29 +122,119 @@ proc snapshotListeners*(
...
+  listeners
 
-# ---------------------------------------------------------------------------
-# Dispatch templates (used by {.ffiEvent.}-generated procs)
-# ---------------------------------------------------------------------------
 
-var ffiCurrentEventRegistry* {.threadvar.}: ptr FFIEventRegistry
-  ## Set by the FFI thread at startup so dispatchFFIEvent / dispatchFFIEventCbor
-  ## can find their registry without taking a context pointer per call site.
+const EventQueueCapacity* = 1024
```
Ivansete-status, line 133: Maybe is interesting to create another lib event `EventQueueAlmostFull` or something like that when the queue len is approaching 80% capacity. In which case, the event should go first and not be last. Similarly, the `NotResponding` event should also always go first in the queue, i.e., top priority.

ffi/ffi_events.nim
```diff
@@ -179,53 +249,39 @@ template withFFIEventDispatch(
...
-    for listener in listeners:
-      listener.callback(
-        RET_OK,
-        cast[ptr cchar](unsafeAddr event[0]),
-        cast[csize_t](len(event)),
-        listener.userData,
-      )
+    let dataPtr: pointer =
+      if event.len > 0: unsafeAddr event[0]
+      else: nil
```
Ivansete-status: This snippet is used many times. Perhaps we can encapsulate it.

ffi/ffi_events.nim
```diff
@@ -142,29 +122,119 @@ proc snapshotListeners*(
...
+
+proc notifyListenersErr*(listeners: seq[FFIEventListener], msg: string) =
+  ## Fans out an error message to every listener in the snapshot.
+  let dataPtr =
+    if msg.len > 0: cast[ptr cchar](unsafeAddr msg[0])
+    else: nil
+  for listener in listeners:
+    listener.callback(
+      RET_ERR, dataPtr, cast[csize_t](msg.len), listener.userData
+    )
```
Ivansete-status: Maybe we can adjust these two a little. It sounds interesting to keep a generic `notifyListeners` because maybe we want to encode a CBOR object (or C struct) in an error event.

```suggestion

const emptyListenerPayload: cstring = ""
  ## Non-nil zero-length buffer handed to listeners when a payload is
  ## empty, so a consumer doing `std::string(data, len)` / `memcpy` never
  ## receives a nil pointer (which is UB even at len 0).

proc notifyListeners*(
    listeners: seq[FFIEventListener], retCode: cint, data: pointer, dataLen: int
) =
  let n = max(dataLen, 0)
  let dataPtr =
    if n > 0 and not data.isNil(): cast[ptr cchar](data)
    else: cast[ptr cchar](emptyListenerPayload)
  for listener in listeners:
    listener.callback(retCode, dataPtr, cast[csize_t](n), listener.userData)

proc notifyListenersErr*(listeners: seq[FFIEventListener], msg: string) =
  ## Error fan-out: adapts the message string to `notifyListeners`, which
  ## supplies the non-nil pointer for the empty-message case.
  let p =
    if msg.len > 0: cast[pointer](unsafeAddr msg[0])
    else: cast[pointer](emptyListenerPayload)
  notifyListeners(listeners, RET_ERR, p, msg.len)
```

Ivansete-status (commented):
Some comments so far, thanks for it!
The main point that needs to be addressed is to make sure the events' callbacks are indeed handled by the event-thread so that the ffi-thread cannot be blocked.

ffi/ffi_context.nim
gmelodie: I think this is implemented in #69. I didn't want this PR to grow much more than what it is already.

ffi/ffi_events.nim
gmelodie: Maybe we could tackle this in #69 or in another PR?

ffi/ffi_context.nim
```diff
@@ -347,33 +165,41 @@ proc initContextResources*[T](ctx: ptr FFIContext[T]): Result[void, string] =
...
   if not stopSignaled:
     return err("failed to signal stopSignal on time in signalStop")
-  return ok()
+  # Non-fatal: event thread will see running==false on the next tick.
+  let evtSignaled = ctx.eventQueueSignal.fireSync()
+  if evtSignaled.isErr():
+    error "failed to signal eventQueueSignal in signalStop", error = evtSignaled.error
+  elif evtSignaled.get() == false:
+    error "failed to signal eventQueueSignal on time in signalStop"
+  ok()
```
Ivansete-status: nit, I'd also leave restyle for separate PR.
```suggestion
  return ok()
```

ffi/ffi_events.nim
Ivansete-status: Yes, either option is fine to me as long as we have it :)

ffi/ffi_events.nim
```diff
@@ -142,29 +122,119 @@ proc snapshotListeners*(
...
+    if n > 0 and not data.isNil(): cast[ptr cchar](data)
+    else: cast[ptr cchar](emptyListenerPayload)
+  for listener in listeners:
+    listener.callback(retCode, dataPtr, cast[csize_t](n), listener.userData)
+
+proc notifyListenersErr*(listeners: seq[FFIEventListener], msg: string) =
+  ## Error fan-out: adapts the message string to `notifyListeners`, which
+  ## supplies the non-nil pointer for the empty-message case.
+  let p =
+    if msg.len > 0: cast[pointer](unsafeAddr msg[0]) else: nil
```
Ivansete-status: I vote to always avoid passing nil to any callback result. We can pass the `emptyListenerPayload` in this case. Applies elsewhere.

Ivansete-status (approved):
LGTM! Thanks for it! 🙌 
Just added minor nits

PR #74, opened 2026-06-04
fix(pool): recycle contexts to stop per-cycle fd leak

Ivansete-status:
destroyFFIContext stopped and joined the worker threads on every release, and createFFIContext rebuilt them on the next acquire. Each cycle therefore allocated a fresh worker — 4 ThreadSignalPtr socketpairs + the ffi/watchdog thread chronos dispatcher kqueues — and never reclaimed the old ones (the pool deliberately skips closing them, relying on slot reuse that never actually reused the resources). A consumer that creates and destroys contexts repeatedly (e.g. nim-sds ReliabilityManager) leaked ~10 fds per cycle, unbounded.

Make the pool genuinely reuse a slot's worker:

- Track per-slot `initialized`; createFFIContext builds the worker once and reuses it on every later acquisition of the same slot.
- Add releaseFFIContext: parks a context (returns its slot) WITHOUT stopping the threads, so the next acquire reuses the same fds. It also drops the stale C event callback so a watchdog tick on a parked slot cannot invoke a callback whose user-data the consumer may already have freed. The caller is responsible for quiescing its library object (on the FFI thread) first.
- destroyFFIContext keeps full-teardown semantics for error/non-pooling paths and now marks the slot uninitialised so a later acquire rebuilds it.

Tests: add a park & reuse suite (same-slot live-worker reuse, callback/lib pointer dropped on park, and fd usage bounded across 20 park/reuse cycles). The fd test fails by ~10 fds/cycle against the pre-fix behaviour. Green under both --mm:refc and --mm:orc.

### DIFF
--- a/ffi/ffi_context.nim
+++ b/ffi/ffi_context.nim
@@ -361,7 +361,7 @@ proc signalStop*[T](ctx: ptr FFIContext[T]): Result[void, string] =
 
 ## If the FFI thread's event loop is blocked by a synchronous handler
 ## (e.g. blocking I/O), it cannot process reqSignal in time to exit.
-## clearContext waits on threadExitSignal up to this bound; on timeout it
+## stopAndJoinThreads waits on threadExitSignal up to this bound; on timeout it
 ## returns err and skips joinThread/cleanup (leaking the thread + ctx slot)
 ## rather than hanging the caller forever.
 const ThreadExitTimeout* = 1500.milliseconds
@@ -384,12 +384,4 @@ proc stopAndJoinThreads*[T](ctx: ptr FFIContext[T]): Result[void, string] =
 
   joinThread(ctx.ffiThread)
   joinThread(ctx.watchdogThread)
-  return ok()
-
-proc clearContext[T](ctx: ptr FFIContext[T]): Result[void, string] =
-  ## Stops the FFI context that was created via createFFIContext[T]() (heap).
-  ctx.stopAndJoinThreads().isOkOr:
-    return err("clearContext: " & $error)
-  ctx.cleanUpResources().isOkOr:
-    return err("cleanUpResources failed: " & $error)
   return ok()
--- a/ffi/ffi_context.nim
+++ b/ffi/ffi_context.nim
@@ -14,7 +14,7 @@ type FFICallbackState* = object
 
 type CtxLifecycle {.pure.} = enum
   ## Request-acceptance + recycle handshake for a pooled context, held as an
-  ## Atomic on FFIContext. ("Recycle" = drain the context and return its slot to
+  ## Atomic on FFIContext. ("Recycle" = drain the context and return its context to
   ## the pool for reuse, keeping the worker alive — unlike destroyFFIContext, which
   ## fully tears the threads down.) Invariants:
   ##   * Requests are accepted ONLY in `Active`; the gate in sendRequestToFFIThread
@@ -23,7 +23,7 @@ type CtxLifecycle {.pure.} = enum
   ##       Active         -> RecyclePending   requestRecycle (caller, under `lock`)
   ##       RecyclePending -> Recycling        the FFI loop   (one-shot compareExchange)
   ##       Recycling      -> Active           markReacquired (caller, on reuse)
-  ##     (initContextResources starts a slot in `Active`.)
+  ##     (initContextResources starts a context in `Active`.)
   ##   * The gate stays closed across BOTH RecyclePending and Recycling, so no
   ##     request can dispatch onto a context being recycled or about to be reused.
   ##   * Only the FFI loop makes the RecyclePending -> Recycling move, so the
@@ -61,8 +61,8 @@ type FFIContext*[T] = object
     # RET_OK once drained, RET_ERR if it timed out. Set by requestRecycle.
   recycleUserData: pointer
   inUse: Atomic[bool]
-    # Whether the slot is claimed. createFFIContext claims it (false -> true); the
-    # recycle handler clears it once drained. On the slot so the owning thread can
+    # Whether the context is claimed. createFFIContext claims it (false -> true); the
+    # recycle handler clears it once drained. On the context so the owning thread can
     # release it without reaching into the pool.
   registeredRequests: ptr Table[cstring, FFIRequestProc]
     # Pointer to with the registered requests at compile time
@@ -129,7 +129,7 @@ proc sendRequestToFFIThread*(
     ctx.lock.release()
 
   ## A recycle closes this gate (under the same lock), so a queued or late sender
-  ## bails here instead of dispatching onto a slot about to be reused.
+  ## bails here instead of dispatching onto a context about to be reused.
   if ctx.lifecycle.load() != CtxLifecycle.Active:
     deleteRequest(ffiRequest)
     return err("FFI context is not accepting requests (being recycled)")
@@ -293,7 +293,7 @@ proc recycleContext[T](
     ctx: ptr FFIContext[T], pending: ptr seq[Future[void]]
 ) {.async.} =
   ## Recycle handler, on the FFI thread (requestRecycle already closed the gate):
-  ## drain the in-flight handlers, free the lib object, release the slot for reuse,
+  ## drain the in-flight handlers, free the lib object, release the context for reuse,
   ## and fire the callback with the outcome. Never blocks the caller.
   ##
   ## `pending` is the run loop's seq of handler Futures, by ptr (async procs can't
@@ -306,7 +306,7 @@ proc recycleContext[T](
     naturallyDrained = await allFutures(pending[]).withTimeout(RecycleTimeout)
 
   ## 2. If any are wedged, cancel them and give the cancellations a bounded moment
-  ##    to unwind, so the slot can be reclaimed rather than leaked.
+  ##    to unwind, so the context can be reclaimed rather than leaked.
   var safeToRecycle = naturallyDrained
   if not naturallyDrained:
     for fut in pending[]:
@@ -319,7 +319,7 @@ proc recycleContext[T](
   ctx.recycleCallback = nil
 
   if safeToRecycle:
-    ## Nothing can touch the context now. Free the lib here, then release the slot
+    ## Nothing can touch the context now. Free the lib here, then release the context
     ## BEFORE the callback (the atomic store publishes these writes to whoever
     ## reclaims it) so a caller reacquiring on the callback finds it already free.
     freeLib(ctx)
@@ -361,7 +361,7 @@ proc ffiThreadBody[T](ctx: ptr FFIContext[T]) {.thread.} =
     while ctx.running.load():
       ## Recycle requested: claim it (RecyclePending -> Recycling, one-shot) and run
       ## the recycle on this owning thread, then keep looping so the worker stays
-      ## alive for the slot's next reuse.
+      ## alive for the context's next reuse.
       var expected = CtxLifecycle.RecyclePending
       if ctx.lifecycle.compareExchange(expected, CtxLifecycle.Recycling):
         await recycleContext(ctx, addr pending)
@@ -425,9 +425,9 @@ proc cleanUpResources[T](ctx: ptr FFIContext[T]): Result[void, string] =
   return ok()
 
 proc initContextResources*[T](ctx: ptr FFIContext[T]): Result[void, string] =
-  ## Initialises all resources inside an already-allocated FFIContext slot.
+  ## Initialises all resources inside an already-allocated FFIContext.
   ## On failure every partially-initialised resource is closed; the caller
-  ## is responsible for releasing the slot (freeShared or pool.releaseSlot).
+  ## is responsible for releasing the context (freeShared or ctx.unclaim()).
   ctx.lock.initLock()
 
   var success = false
@@ -491,7 +491,7 @@ proc signalStop*[T](ctx: ptr FFIContext[T]): Result[void, string] =
 ## If the FFI thread's event loop is blocked by a synchronous handler
 ## (e.g. blocking I/O), it cannot process reqSignal in time to exit.
 ## stopAndJoinThreads waits on threadExitSignal up to this bound; on timeout it
-## returns err and skips joinThread/cleanup (leaking the thread + ctx slot)
+## returns err and skips joinThread/cleanup (leaking the thread + ctx)
 ## rather than hanging the caller forever.
 const ThreadExitTimeout* = 1500.milliseconds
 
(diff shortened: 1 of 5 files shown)


ffi/ffi_context_pool.nim
```diff
@@ -30,24 +28,55 @@ proc releaseSlot[T](pool: var FFIContextPool[T], ctx: ptr FFIContext[T]) =
...
+  ## Parks a context for reuse: returns its slot to the pool WITHOUT stopping the
+  ## worker threads, so the next createFFIContext reuses the same fds. The caller
+  ## MUST have already quiesced its library object (e.g. cancelled the object's
+  ## async tasks) on the FFI thread before calling this — the worker keeps
+  ## running. The stale C event callback is dropped so a watchdog tick on the
+  ## parked slot cannot invoke a callback whose user-data may already be freed.
+  ctx.callbackState = default(FFICallbackState)
+  ctx.myLib = nil
+  pool.releaseSlot(ctx)
+  return ok()
```
NagyZoltanPeter: On which thread it is running?
I'm not sure if events are moved onto delivery thread this can be that simple and not wait for draining all in-flight (under processing) events.
I may not see the whole pic here:
- who and when tells to myLib to tear down (if not mistaken that's the master object of the lib context)?
- How the user can be sure that there are no more in-flight processing?

tests/test_ffi_context.nim
```diff
@@ -648,3 +648,109 @@ suite "ptr return type in .ffi.":
...
+    if findExe("lsof").len == 0:
+      return -1
+    try:
+      let output =
+        execProcess("lsof", args = ["-p", $getCurrentProcessId()], options = {poUsePath})
+      return output.splitLines().countIt(it.len > 0)
+    except CatchableError:
+      return -1
+
+suite "releaseFFIContext (park & reuse)":
```
NagyZoltanPeter, line 811: While all these ok, I miss the test of these from real FFI use? A c++ / rust way that not just spin-up and release the context but does this during heavy load.

ffi/ffi_context_pool.nim
```diff
@@ -30,24 +28,55 @@ proc releaseSlot[T](pool: var FFIContextPool[T], ctx: ptr FFIContext[T]) =
...
+      ## Reused slot: worker threads, dispatcher and signals are already alive.
+      return ok(ctx)
+    initContextResources(ctx).isOkOr:
+      pool.inUse[i].store(false)
+      return err("createFFIContext: initContextResources failed: " & $error)
+    pool.initialized[i].store(true)
+    return ok(ctx)
+  return err("FFI context pool exhausted (max " & $MaxFFIContexts & " contexts)")
+
+proc releaseFFIContext*[T](
```
NagyZoltanPeter, line 33: Is it wired into some exported functions too?

NagyZoltanPeter (commented):
I leave some questions here before able to understand what exactly I am looking at.

tests/test_ffi_context.nim
Ivansete-status: The issue has been detected by an existing test in status-go.
I've validated that these tests enforces the leak issue (8 fds for thread signal ptr and 2 fds per kqueue)
There's no need to further tests because these already cover the fix we want to cover.

ffi/ffi_context_pool.nim
Ivansete-status: That was invoked by the caller's thread. Addressed in 1c99eaa1898b039229f1946a051a09e5d26b2ad5

ffi/ffi_context_pool.nim
Ivansete-status: Good catch thanks! Addressed in a combination of 1c99eaa1898b039229f1946a051a09e5d26b2ad5 and e43362464565d663b8a6749db252e948496d20cb

tests/test_ffi_context.nim
NagyZoltanPeter: So you are convinced that calling releaseFFIContext with ongoing event handlers for example will be the same from Go?

ffi/ffi_context_pool.nim
NagyZoltanPeter: I see now it in comments. But than it is kind of dangerous or it can be.
Let suppose the user issues some request with async scheduling. The FFI can only handle one at a time, there is no queue behind. So the queue is now represented by a stack of waiting threads on the caller's side.
It means it can concurrently call this releaseFFIContext with other request in 'queue' from other threads.
What happens is such request lands on a context already having nil myLib? Does the FFI call side of the request has a copy of the context already before started waiting?

tests/test_ffi_context.nim
Ivansete-status: Yes, good point. Not really. Then, we will address that in the following: https://github.com/logos-messaging/nim-ffi/issues/75
Thanks!

ffi/internal/ffi_macro.nim
```diff
@@ -1565,18 +1565,19 @@ macro ffiDtor*(prc: untyped): untyped =
     ffiBody.add(bodyNode)
 
   let poolIdent = ident($libTypeName & "FFIPool")
+  # Hand teardown to the FFI thread (releaseFFIContext -> requestRecycle), recycling
+  # the slot for reuse to keep fd usage bounded. NON-BLOCKING: returns RET_OK once
+  # accepted; the real outcome arrives via `callback`, so callers must wait on it,
+  # not the return code.
```
gmelodie:
```suggestion
```

And put this part inside `releaseFFIContext`:
```nim
  ## Recycle the slot for reuse to keep fd usage bounded.
  ## NON-BLOCKING: returns RET_OK once accepted;
  ## the real outcome arrives via `callback`
```
could even leave the callback part out though, as pretty much every callback-based function works this way

ffi/ffi_context_pool.nim
```diff
@@ -12,43 +12,67 @@ type FFIContextPool*[T] = object
...
+      return ok(ctx)
+    initContextResources(ctx).isOkOr:
+      ctx.unclaim()
+      return err("createFFIContext: initContextResources failed: " & $error)
+    pool.initialized[i].store(true)
+    return ok(ctx)
+  return err("FFI context pool exhausted (max " & $MaxFFIContexts & " contexts)")
+
+proc releaseFFIContext*[T](
+    pool: var FFIContextPool[T],
```
gmelodie: Doesn't use the `pool` it seems?
```suggestion
```

ffi/ffi_context_pool.nim
```diff
@@ -12,43 +12,67 @@ type FFIContextPool*[T] = object
...
+    callback: FFICallBack,
+    userData: pointer,
+): Result[void, string] =
+  ## Parks a context for reuse without stopping its worker, so the next
+  ## createFFIContext reuses the same threads and fds. Steady-state cleanup path
+  ## for the generated destructor; destroyFFIContext is for failure/non-pool use.
+  ##
+  ## NON-BLOCKING: the FFI thread drains the handlers, frees the lib and releases
+  ## the slot, then fires `callback` (RET_OK drained, RET_ERR stuck). The slot
+  ## returns to the pool from that thread, so a reused slot never carries a straggler.
```
gmelodie: This docstring should be in `requestRecycle`, not here

ffi/ffi_context_pool.nim
```diff
@@ -12,43 +12,67 @@ type FFIContextPool*[T] = object
...
-  ## Stops the FFI context and returns its slot to the pool. If the FFI thread
-  ## is blocked and does not exit in time, the slot is leaked rather than
-  ## reclaimed — closing its resources while the thread is still live would be
-  ## unsafe.
+  ## Full teardown: stops/joins the worker threads and returns the slot to the
+  ## pool, marking it uninitialised so a later createFFIContext rebuilds it. Used
+  ## on creation failure and by non-pooling callers; steady-state cleanup should
+  ## use releaseFFIContext to keep fd usage bounded. If the FFI thread is blocked
+  ## and does not exit in time, the slot is leaked rather than reclaimed —
+  ## closing its resources while the thread is still live would be unsafe.
```
gmelodie: These long comments... why does claude not just write a book already?
```suggestion
  ## Teardown worker threads and returns the slots to the pool, 
  ## marking as uninitialised so a later createFFIContext rebuilds it.
  ## Steady-state cleanup should use releaseFFIContext to keep fd usage bounded.
  ## If the FFI thread is blocked and does not exit in time, the slot is leaked.
  ## Closing resources while the FFI thread is still live is unsafe.
```

ffi/ffi_context.nim
```diff
@@ -12,6 +12,26 @@ type FFICallbackState* = object
...
+  ##       RecyclePending -> Recycling        the FFI loop   (one-shot compareExchange)
+  ##       Recycling      -> Active           markReacquired (caller, on reuse)
+  ##     (initContextResources starts a slot in `Active`.)
+  ##   * The gate stays closed across BOTH RecyclePending and Recycling, so no
+  ##     request can dispatch onto a context being recycled or about to be reused.
+  ##   * Only the FFI loop makes the RecyclePending -> Recycling move, so the
+  ##     recycle runs exactly once per request.
+  Active            ## serving requests
+  RecyclePending    ## recycle requested; FFI loop hasn't claimed it yet
+  Recycling         ## FFI loop claimed it: draining handlers, then freeing
```
```suggestion
type CtxLifecycle {.pure.} = enum
  ## State machine guarding a pooled FFI context, held as an Atomic on FFIContext.
  ## "Recycle" = drain the context and return its slot to the pool for reuse,
  ## keeping the worker alive (unlike destroyFFIContext, which tears threads down).
  ##
  ## Requests are accepted ONLY in `Active`. The gate stays shut across both
  ## RecyclePending and Recycling, so nothing dispatches onto a context mid-recycle.
  ##
  ## Transitions (who performs them):
  ##   Active         -> RecyclePending   requestRecycle  (caller, under `lock`)
  ##   RecyclePending -> Recycling        FFI loop        (one-shot compareExchange)
  ##   Recycling      -> Active           markReacquired  (caller, on reuse)
  ## Only the FFI loop does RecyclePending -> Recycling, so recycle runs exactly once.
  Active            ## serving requests
  RecyclePending    ## recycle requested; FFI loop hasn't claimed it yet
  Recycling         ## FFI loop claimed it: draining handlers, then freeing
```

ffi/ffi_context.nim
```diff
@@ -33,6 +53,17 @@ type FFIContext*[T] = object
...
+    # Request gate + recycle handshake in one. See CtxLifecycle for the states,
+    # transitions and invariants.
+  recycleCallback: FFICallBack
+    # The destructor's callback, fired by the recycle handler with the outcome:
+    # RET_OK once drained, RET_ERR if it timed out. Set by requestRecycle.
+  recycleUserData: pointer
+  inUse: Atomic[bool]
+    # Whether the slot is claimed. createFFIContext claims it (false -> true); the
+    # recycle handler clears it once drained. On the slot so the owning thread can
+    # release it without reaching into the pool.
```
gmelodie, line 54:
```suggestion
lifecycle: Atomic[CtxLifecycle]
    # Request gate + recycle handshake. See CtxLifecycle for states/transitions.
  recycleCallback: FFICallBack
    # Fired by the recycle handler with the outcome: RET_OK once drained,
    # RET_ERR on timeout. Set by requestRecycle.
  recycleUserData: pointer
  inUse: Atomic[bool]
    # Whether the slot is claimed. createFFIContext sets false -> true; the recycle
    # handler clears it once drained. Lives on the slot so the owning thread can
    # release it without touching the pool.
```

gmelodie (commented):
Half of this PR is comments trying to explain the code, it's just too painful to follow it all. Can we try to get rid of the comments by having better, self-explanatory code? pretty much every line of code is paired with a couple lines of comments explaining it. Even worse, the explanations are convoluted.

NagyZoltanPeter (approved):
I did not find any further threats. LGTM!

ffi/internal/ffi_macro.nim
Ivansete-status: Addressed in 5f76d1fa5b85c056ffda1e1d26d432bbaf0a6c79

ffi/ffi_context_pool.nim
Ivansete-status: Thanks! Addressed in 6c4bbe01cd625d86b8e77bf8e670f0ef7c31758c

ffi/ffi_context_pool.nim
Ivansete-status: Addressed in 1667398f07884ee77f361142f489969c9cbe9dd1

ffi/ffi_context.nim
Ivansete-status: I think is addressed in https://github.com/logos-messaging/nim-ffi/commit/d05f7f6441b745efc68e9880f2cae3cdd0a2602f

ffi/ffi_context.nim
Ivansete-status: I think addressed in d05f7f6441b745efc68e9880f2cae3cdd0a2602f

ffi/internal/ffi_macro.nim
```diff
@@ -1565,18 +1565,19 @@ macro ffiDtor*(prc: untyped): untyped =
...
-    let `destroyResIdent` =
-      `poolIdent`.destroyFFIContext(cast[ptr FFIContext[`libTypeName`]](ctx))
+    let `destroyResIdent` = `poolIdent`.releaseFFIContext(
+      cast[ptr FFIContext[`libTypeName`]](ctx), callback, userData
+    )
     if `destroyResIdent`.isErr():
       if not callback.isNil:
         let errStr = "destroy failed: " & $`destroyResIdent`.error
         callback(RET_ERR, unsafeAddr errStr[0], cast[csize_t](errStr.len), userData)
       return RET_ERR
```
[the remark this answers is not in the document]
Ivansete-status: Thanks! Addressed in c749d890abf6e32dcb0b72b69f37c4740093e062

ffi/ffi_context.nim
```diff
@@ -386,10 +515,44 @@ proc stopAndJoinThreads*[T](ctx: ptr FFIContext[T]): Result[void, string] =
...
+  ## outcome (RET_OK drained, RET_ERR stuck). NON-BLOCKING.
+  ##
+  ## Order matters: set the callback before flipping to RecyclePending (the flip is
+  ## the trigger), under `lock` to serialise the gate with sendRequestToFFIThread.
+  ctx.recycleCallback = callback
+  ctx.recycleUserData = userData
+
+  ctx.lock.acquire()
+  ctx.lifecycle.store(CtxLifecycle.RecyclePending)
+  ctx.lock.release()
```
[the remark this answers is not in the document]
Ivansete-status: Addressed in f07ef745009817e6944583f6116d93d83354d71c

gmelodie (approved):

Issue #80, opened 2026-06-12
Add `{.ffiHandle.}` to export complex live objects as opaque handles


Ivansete-status:
## Problem

An `.ffi.` proc can only return **CBOR-serializable** types. That excludes complex live objects — exactly what callers want a reference to.

`Waku` ([`waku/factory/waku.nim`](../logos-delivery/waku/factory/waku.nim#L59)) is a `ref object` holding `Future[void]` loop handles, an `rng`, a `WakuNode`, REST/metrics servers, a discovery service. None of it is value-copyable, so CBOR is impossible for it — and the same holds for any sub-object (`Kernel`, a session, …).

Today only the singleton `ctx.myLib` crosses, auto-bound as every proc's first param. There's no way to return a second complex object (`waku.kernel()`) or pass one back as a receiver (`kernel.send(...)`). We want:

```nim
kernel = ctx.kernel()
kernel.send(...)        # or: ctx.kernel().send(...)
```

## Proposal: `{.ffiHandle.}` types backed by a per-ctx registry

```nim
type Kernel {.ffiHandle.} = ref object ...
```

- **Wire form is a plain `uint64`.** The object never crosses; only the handle does.
- **The object lives in a per-ctx registry** (`FFIContext.handles`), torn down with the ctx — mirroring `eventRegistry`/`eventQueue`.
- **You mark the type once, not each proc.** Once `Kernel` is `{.ffiHandle.}`, *every* proc that returns it sends a handle id, and *every* proc that takes it receives one — automatically, with no extra annotation at the call site. So there's never a per-proc choice between "send the whole object" and "send a handle": the live object physically can't cross, so the handle is the only representation. A given type is either a value type (`{.ffi.}`, copied) or a handle type (`{.ffiHandle.}`, referenced) — never both.

### Registry: monotonic id + table

Mirror the existing [`FFIEventRegistry`](ffi/ffi_events.nim#L22) idiom — a `nextId` counter plus a table, no slot reuse:

```nim
type FFIHandleRegistry = object
  nextId: uint64                # 0 reserved as null; ids start at 1
  byHandle: Table[uint64, ...]  # handle -> live ref
```

- `register(obj) -> uint64` (`inc nextId; byHandle[nextId] = obj`), `lookup(handle) -> ref|nil`, `release(handle)` (table delete).
- Ids are never recycled, so a stale or forged handle is simply absent → **clean miss → `RET_ERR`**, never a use-after-free.
[...]

PR #81, opened 2026-06-12
feat: {.ffiHandle.} — export complex live objects as opaque handles

gmelodie:
## Summary

Implements #80. Adds `{.ffiHandle.}`, letting an `.ffi.` proc return — or receive — a complex, non-CBOR-serializable live object (a `Waku`, a `Kernel`, a session) by reference instead of by value.

The object never crosses the FFI boundary. Its wire form is a plain `uint64`; the live `ref` lives in a per-ctx registry and is reconstituted on the way back in. A given type is either a value type (`{.ffi.}`, copied) or a handle type (`{.ffiHandle.}`, referenced) — never both — so there is no per-proc "send the object vs send a handle" choice.

```nim
type Kernel {.ffiHandle.} = ref object
  ...

proc demoKernel*(d: Demo, req: MakeReq): Future[Result[Kernel, string]] {.ffi.}   # returns a handle
proc demoKernelSend*(k: Kernel, msg: string): Future[Result[string, string]] {.ffi.}  # handle is the receiver
```

## Design

- **`ffi/ffi_handles.nim`** — `FFIHandleRegistry`, mirroring the `FFIEventRegistry` idiom: monotonic `nextId` (0 = null, ids never recycled) + a `Table`. `register` / `lookup` / `release` / `releaseAll`. Each entry records its type name, so a stale, forged, or wrong-type id misses cleanly → `RET_ERR`, never a use-after-free. Single-threaded (FFI-thread) access, so no locking. Handle types inherit an injected `FFIHandleRoot` base so heterogeneous refs are storable under one static type.
- **`FFIContext`** — new `handles` field, init/deinit alongside `eventRegistry`. Refs are freed on the FFI thread (`ffiThreadBody` `defer`) because refc heaps are thread-local.
- **Macro forks in `ffi/internal/ffi_macro.nim`:**
  - `ffiHandle` type macro — validates `ref object`, injects `of FFIHandleRoot`, records the name.
  - `storageType` — a handle param stores as `uint64` in the per-proc Req.
  - Return path — a handle return is `register`ed; only the id is CBOR-encoded.
  - Param path — a handle param is `lookup`ed, returning `RET_ERR` on a miss.
  - First-param (receiver) fork — a handle receiver resolves `FFIContext[LibType]` / pool from the lib type recorded by `declareLibrary`.
  - Composes with `{.ffiRaw.}` with zero special-casing: the wire type is already `uint64`, so raw bodies call `ctx.handles.lookup/register/release` by hand.
- **Codegen (rust / cpp / cddl)** — handle params and returns map to `u64` / `uint64_t` / `uint`; no struct is emitted for a handle type. The flat `kernel_send(ctx, handle, …)` form works; the optional OO/RAII wrapper class is left for a follow-up.
- **`declareLibrary`** now also declares the `<LibType>FFIPool` var (previously only `ffiCtor`/`ffiDtor` did) and records the lib type name.

## Testing

- `tests/unit/test_ffi_handle.nim`: handle round-trip (out as `uint64` → reconstituted live object), handle-as-receiver, and forged / null id → `RET_ERR`. Passes under **orc** and **refc**, and is **ASAN/UBSan-clean**.
- Existing suites pass; the timer/echo example bindings regenerate unchanged (no codegen regression); `nph --check` clean.

### DIFF
--- a/ffi/ffi_handles.nim
+++ b/ffi/ffi_handles.nim
@@ -4,6 +4,8 @@
 ## FFI-thread-only access, so no locking.
 
 import std/tables
+import results
+import ./cbor_serial
 
 type
   FFIHandleRoot* = ref object of RootObj
@@ -36,20 +38,29 @@ proc register*(
 
 proc lookup*(
     reg: var FFIHandleRegistry, handle: uint64, typeName: string
-): FFIHandleRoot =
-  ## Live ref for `handle`, or nil if absent or registered under another type.
+): Result[FFIHandleRoot, string] =
+  ## Live ref for `handle`; err if absent or registered under another type.
   let entry = reg.byHandle.getOrDefault(handle)
-  if entry.obj.isNil() or entry.typeName != typeName:
-    return nil
-  entry.obj
+  if entry.obj.isNil():
+    return err("no ffiHandle with id " & $handle)
+  if entry.typeName != typeName:
+    return err(
+      "ffiHandle " & $handle & " has type '" & entry.typeName & "', expected '" &
+        typeName & "'"
+    )
+  ok(entry.obj)
 
 proc release*(reg: var FFIHandleRegistry, handle: uint64): bool {.discardable.} =
   ## Drops the entry; true iff it existed.
   if not reg.byHandle.hasKey(handle):
     return false
   reg.byHandle.del(handle)
-  true
+  return true
 
 proc releaseAll*(reg: var FFIHandleRegistry) =
   ## Drops every entry. Must run on the FFI thread that allocated the refs.
   reg.byHandle.clear()
+
+proc encodeHandle*(id: uint64): seq[byte] =
+  ## Wire-form bytes for a handle id. Single ABI seam for future format changes.
+  cborEncode(id)
--- a/ffi/ffi_handles.nim
+++ b/ffi/ffi_handles.nim
@@ -1,0 +1,55 @@
+## Per-context registry of live `{.ffiHandle.}` objects. The object stays here,
+## in `FFIContext.handles`; only its `uint64` id crosses the boundary. Ids are
+## monotonic and never recycled (0 = null), so a stale/forged id misses cleanly.
+## FFI-thread-only access, so no locking.
+
+import std/tables
+
+type
+  FFIHandleRoot* = ref object of RootObj
+    ## Base every `{.ffiHandle.}` type inherits from, so handle refs are storable
+    ## under one static type.
+
+  FFIHandleEntry = object
+    obj: FFIHandleRoot
+    typeName: string
+
+  FFIHandleRegistry* = object
+    nextId*: uint64
+    byHandle*: Table[uint64, FFIHandleEntry]
+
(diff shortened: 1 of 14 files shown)


ffi/ffi_handles.nim
```diff
@@ -0,0 +1,55 @@
...
+    reg: var FFIHandleRegistry, obj: FFIHandleRoot, typeName: string
+): uint64 =
+  ## Stores `obj`, returns its fresh handle id (>0).
+  reg.nextId.inc()
+  reg.byHandle[reg.nextId] = FFIHandleEntry(obj: obj, typeName: typeName)
+  reg.nextId
+
+proc lookup*(
+    reg: var FFIHandleRegistry, handle: uint64, typeName: string
+): FFIHandleRoot =
```
Ivansete-status: Maybe better to return a `Result`?

ffi/internal/ffi_macro.nim
```diff
@@ -385,23 +409,34 @@ proc addNewRequestToRegistry(reqTypeName, reqHandler: NimNode): NimNode =
...
     let `typedResIdent` = await `callExpr`
     if `typedResIdent`.isErr:
       return err(`typedResIdent`.error)
     when typeof(`typedResIdent`.value) is seq[byte]:
       return ok(`typedResIdent`.value)
     elif typeof(`typedResIdent`.value) is void:
       return ok(newSeq[byte]())
+    elif typeof(`typedResIdent`.value) is FFIHandleRoot:
+      return ok(
+        cborEncode(
```
Ivansete-status: It seems a little bit hard-coded. In upcoming PRs we might need to be able to adapt its ABI format.

ffi/internal/ffi_macro.nim
```diff
@@ -631,8 +700,20 @@ macro ffi*(prc: untyped): untyped =
     error("`.ffi.` procs require at least 1 parameter (the library type)")
 
   let firstParam = formalParams[1]
-  let libParamName = firstParam[0]
-  let libTypeName = firstParam[1]
+  let recvName = firstParam[0]
+  let recvType = firstParam[1]
+  let firstIsHandle = isHandleType(recvType)
+  if firstIsHandle and currentLibType.len == 0:
```
Ivansete-status, line 705: I wonder if we could force, at compile time, that any time annotated with `.ffiHandle.`, can only be passed as first param in an `ffi` proc. Disregard that if this is too complex.

ffi/ffi_handles.nim
```diff
@@ -0,0 +1,55 @@
...
+  if entry.obj.isNil() or entry.typeName != typeName:
+    return nil
+  entry.obj
+
+proc release*(reg: var FFIHandleRegistry, handle: uint64): bool {.discardable.} =
+  ## Drops the entry; true iff it existed.
+  if not reg.byHandle.hasKey(handle):
+    return false
+  reg.byHandle.del(handle)
+  true
```
Ivansete-status: nit, I like to always be explicit,
```suggestion
  return true
```

Ivansete-status (approved):
LGTM! Thanks for it! 💯

Issue #78, opened 2026-06-09
`ffi`/`ffiEvent` and `ffiRaw`/`ffiEventRaw`


gmelodie:
Instead of relying on compilation flags or pragma options, we should have a mechanism to indicate the ABI format being used on each FFI interaction.

We need to adapt the `declareLibrary` macro so that it accepts something like:

```nim
declareLibrary("libname", LibName, defaultABIFormat = "cbor")
```

With that, any type or proc annotate with `.ffi.`, will use the `cbor`. (the other possible ABI format it `c`)

Also, it is possible to override that default behavior on each particular case.
For that, the following macros...
- `.ffi.`
- `.ffiHandler.`
- `.ffiEvent.`
- etc

...can get their default ABI format overridden by:

- `.ffi: "abi = c".` or `.ffi: "abi = cbor".`
- `.ffiHandler: "abi = c".` or `.ffiHandler: "abi = cbor".`
- `.ffiEvent: "abi = c".` or `.ffiEvent: "abi = cbor".`
- etc

gmelodie:
Implemented in #85 #86 #87 and #88

PR #85, opened 2026-06-22
feat(ffi): per-interaction ABI-format annotations + c codec (ffiRaw)

gmelodie:
## Summary

Implements the per-interaction ABI-format mechanism requested in #78. Instead of relying on compile flags or pragma options, the wire format is now declared on the library and overridable on each FFI interaction.

- `declareLibrary("name", LibType, defaultABIFormat = "cbor")` sets the library-wide default (`"cbor"` or `"c"`) that every FFI annotation inherits.
- Each `{.ffi.}` / `{.ffiRaw.}` / `{.ffiCtor.}` / `{.ffiDtor.}` / `{.ffiEvent.}` / `{.ffiHandle.}` can override it with an `"abi = c"` / `"abi = cbor"` spec (e.g. `{.ffi: "abi = cbor".}`, `{.ffiEvent("on_x", "abi = cbor").}`).
- `declareLibrary` is now required before any FFI annotation, so the name/type/default-ABI are always established first (`requireLibraryDeclared`).

First part of #78.

## How it works

- `ffi/codegen/meta.nim` — `ABIFormat` enum, `parseABIFormatName` / `parseAbiSpec` (case-insensitive, whitespace-tolerant, human-readable errors), the inherited `currentDefaultABIFormat`, and `abiCodegenImplemented` — the single seam that gates formats whose dispatch path isn't wired yet.
- `ffi/internal/ffi_macro.nim` — `resolveABIFormat` (inherit-or-override) + `gateABIFormat`; every annotation records its resolved format on the codegen registry entries.
- `ffi/internal/c_macro_helpers.nim` + `c_wire.nim` — the `c` (flat C-struct) codec: every `{.ffi: "abi = c".}` type gets a `<T>_CWire` companion plus `cwirePack` / `cwireUnpack` / `cwireFree`, covering scalars, strings, `seq`, `Option`, `seq[byte]`, and nested `{.ffi.}` structs.

## Scope / phasing

The `c` codec works end-to-end at the type level, but the `c` proc-dispatch path and the foreign (C++/Rust) generators are still pending. `gateABIFormat` therefore rejects `"abi = c"` on proc/ctor/dtor/event annotations for now, failing loudly rather than emitting CBOR mislabeled as C. Flipping `abiCodegenImplemented` is the one-line change a follow-up makes when `c` dispatch lands.

## Tests

- `tests/unit/test_abi_format.nim` — spec parsing (valid/malformed/unknown), inherited-default vs explicit-override recording on the registry, the `c`-is-gated invariant, and `{.ffiHandle.}` accepting the spec.
- `tests/unit/test_c_wire.nim` — `cwireUnpack(cwirePack(x)) == x` across all field shapes, including empty strings, empty seqs, `none` Options, and `some("")` not collapsing to `none`.
- `tests/bench/bench_codec.nim` (+ `nimble bench_codec`) — single-process microbench isolating `cbor` vs `c` codec cost across payload shapes/sizes.
- CI runs `test_abi_format` and `test_c_wire` across the full OS matrix (`.github/workflows/ci.yml`); existing unit tests gained the `declareLibrary` calls now required.


ffi/codegen/meta.nim
```diff
@@ -30,24 +39,64 @@ type
...
+  of "c":
+    (true, ABIFormat.C)
+  else:
+    (false, ABIFormat.Cbor)
+
+proc parseAbiSpec*(spec: string): tuple[ok: bool, fmt: ABIFormat, err: string] =
+  ## Parse an `"abi = <format>"` override (whitespace/case tolerant). On bad
+  ## grammar or format, returns `ok = false` with a human-readable `err`.
+  let parts = spec.split('=')
+  if parts.len != 2 or parts[0].strip().toLowerAscii() != "abi":
```
Ivansete-status: I think is better to separate those statements vertically to get better insight in case of `false`, to know which branch is `false`.

ffi/internal/ffi_macro.nim
```diff
@@ -7,6 +7,39 @@ when defined(ffiGenBindings):
   import ../codegen/cpp
   import ../codegen/cddl
 
+proc requireLibraryDeclared(where: string) {.compileTime.} =
+  ## Enforce that `declareLibrary(...)` (which records name/type/default-ABI)
+  ## ran before this annotation.
+  if not libraryDeclared:
+    error(
+      where & ": declareLibrary(name, LibType[, defaultABIFormat]) must be " &
+        "called before any FFI annotation"
```
Ivansete-status: nit, maybe better to now cut the line to help future lookups.

ffi/internal/ffi_macro.nim
```diff
@@ -7,6 +7,39 @@ when defined(ffiGenBindings):
...
+      error(parsed.err)
+    fmt = parsed.fmt
+  fmt
+
+proc gateABIFormat(fmt: ABIFormat, where: string) {.compileTime.} =
+  ## Abort if the selected ABI's codegen isn't wired yet (only `Cbor` is), so a
+  ## `c` request fails loudly instead of emitting CBOR mislabeled as C.
+  if not abiCodegenImplemented(fmt):
+    error(
+      where & ": ABI format '" & $fmt & "' is recognized but not yet implemented; " &
```
Ivansete-status: To ease future error lookups is better to leave the variables after a colon.
For example, `"the following ABI format is recognized but not supported: " & $fmt`

tests/unit/test_ctx_validation.nim
```diff
@@ -5,6 +5,11 @@ import ffi
 
 type TestLib = object
 
+# Stub the dylib NimMain importc that declareLibrary emits (this links as a plain exe).
+{.emit: "void libctxvaltestNimMain(void) {}".}
```
Ivansete-status, line 9: Why this is needed?

Ivansete-status (approved):

PR #87, opened 2026-06-22
feat(ffi): cwire codec core (flat scalars + strings)

gmelodie:
## Summary

Second of three PRs splitting the `c` (flat C-struct) ABI codec off the combined #85. This one lands the **codec core**: the flat path only — POD scalars and `string` (as `cstring`). Composite fields (`seq` / `Option` / `seq[byte]` / nested `{.ffi.}`) and the cross-OS CI job follow in PR3.

Stacked on `adopt-abi-mode-annotations` (PR1, #85); this PR is the single `feat: cwire core` commit on top.

## How it works

A `{.ffi: "abi = c".}` object type can't emit its companion at the type site — a type-pragma macro may only return a `TypeDef`. So the macro records the chosen ABI on the registry, and `genBindings()` flushes a `<T>_CWire` companion plus `cwirePack` / `cwireUnpack` / `cwireFree` for every `abiFormat == C` type.

- `ffi/internal/c_macro_helpers.nim` (new) — compile-time codec, flat path: `emittedCWire` tracking, `wireFieldType` / `wireFieldsFor` (`string`→`cstring`, POD unchanged), `buildCWireTypeDef`, `emit{Pack,Unpack,Free}Stmt` (string + POD branches), `buildCWireProcs`, and a non-recursive `ensureCWireFor`.
- `ffi/internal/c_wire.nim` (new) — runtime `cwireAllocStr` / `cwireFreeStr`, routed through `alloc.alloc` / `alloc.dealloc` (malloc/free) to match this branch's cross-thread-safe allocator, not `deallocShared`.
- `ffi/internal/ffi_macro.nim` — re-add `import ./c_macro_helpers`, the `genBindings` cwire flush, and un-gate `c` on `{.ffi.}` *types* now that the codec exists. `c` stays gated on proc/ctor/dtor/event via the unchanged `abiCodegenImplemented` seam.
- `ffi.nim` — export `c_wire`.

## Scope / phasing

`abiCodegenImplemented(C)` is still `false`, so `"abi = c"` remains rejected on proc/ctor/dtor/event annotations — only the codec is wired today. PR3 adds the composite-field branches plus the full `Outer`/`Inner` round-trip test and the cross-OS CI job; a later PR flips the seam to enable `c` proc-dispatch.

## Tests

- `tests/unit/test_c_wire.nim` (new) — flat-struct (`string` + `int` + `bool`) `cwireUnpack(cwirePack(x)) == x` round-trip, including the empty-string edge case.
- Verified locally: full unit suite green under `--mm:orc`, `test_c_wire` green under both `orc` and `refc`; `nph --check` clean; library (`--app:lib`) and timer Rust-binding generation both build.


ffi/internal/c_wire.nim
```diff
@@ -0,0 +1,14 @@
...
+## `cstring` fields need allocation, packed on pack and released on free.
+
+import ../alloc
+
+proc cwireAllocStr*(s: string): cstring {.inline.} =
+  ## NUL-terminated `malloc` copy of `s` (see `ffi/alloc.nim`); pair with
+  ## `cwireFreeStr`. Empty input still yields a valid buffer, never NULL.
+  alloc.alloc(s)
+
+proc cwireFreeStr*(s: cstring) {.inline.} =
```
Ivansete-status: I wonder if we could better protect that against a double-free and pass it by ref and not by value.

ffi/internal/c_macro_helpers.nim
```diff
@@ -0,0 +1,188 @@
...
+
+proc isCWireEmitted(typeName: string): bool {.compileTime.} =
+  ## Indexed scan: works around a Nim 2.2 compile-time VM quirk where `for x in
+  ## seq` over a freshly-mutated `{.compileTime.}` seq goes stale.
+  var found = false
+  for i in 0 ..< emittedCWireTypes.len:
+    if emittedCWireTypes[i] == typeName:
+      found = true
+      break
+  found
```
Ivansete-status: nit, I like to be more explicit.
```suggestion
  return found
```

ffi/internal/c_macro_helpers.nim
```diff
@@ -0,0 +1,188 @@
...
+
+proc markCWireEmitted(typeName: string) {.compileTime.} =
+  if not isCWireEmitted(typeName):
+    emittedCWireTypes.add(typeName)
+
+proc cwireTypeName(userTypeName: string): string =
+  ## Companion-type naming convention; stable so generated tests reach in by name.
+  userTypeName & "_CWire"
+
+proc isStringType(t: NimNode): bool =
```
Ivansete-status, line 26: If we don't support all types with abi = c, maybe we could force it at compile time to avoid possible UB. I mean, make it explicit that only scalars, bool, floats, string are supported.

Also, clearly state at compile time that cbor supports everything and have a clear list of unsupported types in `abi = c`.

And eventually, in future PRs, we can extend the whitelist by adding `seq[T]` or `Opt[T]`, etc.

Ivansete-status (approved):

PR #88, opened 2026-06-22
feat(ffi): cwire codec composite fields

gmelodie:
## Summary

PR3 of the 3-PR `c` (cwire) ABI split. PR2 landed the flat scalar+string codec; this completes it with the **composite-field** branches so any `{.ffi: "abi = c".}` object round-trips through its `*_CWire` companion regardless of field shape — `seq`, `Option`, nested `{.ffi.}` structs, fixed `array`, and named `tuple`, nested to any depth. Stacked on `feat/cwire`, so the diff here is just the composite delta.

Why this is its own PR: the flat path (PR2) is small and self-contained, but composites pull in shared-memory ownership (`allocShared`/`deallocShared`), pointer-as-Option encoding, inline fixed-array packing, and a recursive dependency flush — enough surface to review on its own, and allocator-/layout-sensitive enough to warrant the full-OS CI matrix.

## What's added

In `ffi/internal/c_macro_helpers.nim`, the wire-form field mapping now covers every container shape:

- `seq[T]` → `<name>_items: ptr UncheckedArray[T_w]` + `<name>_len: int`
- `Option[T]` / `Maybe[T]` → `ptr T_w` (nil = none)
- nested `{.ffi.}` `T` → `T_CWire` by value
- `array[N, T]` → inline `array[N, T_w]` (the array itself needs no allocation — only its GC'd element contents do)
- `tuple[a: T, ...]` → `tuple[a: T_w, ...]`

Pack/unpack/free is factored into per-shape emitters (`emitSeq*`, `emitOption*`, `emitArray*`, `emitTuple*`) over a shared element-level core (`emitElem{Pack,Unpack,Free}`), so containers nest recursively — `seq[Option[T]]`, `array[N, tuple[...]]`, `Option[array[...]]`, `array[N, array[M, T]]` all fall out of the same recursion. Allocation is elided where a type owns nothing (`cwireNeedsFree`). The one restriction: a `seq` has no single-field wire form (only the top-level `_items`/`_len` split), so it can't nest inside another container — rejected at compile time with a single clear error.

`collectNestedFFITypes` + the recursive dep-flush in `ensureCWireFor` walk through `seq`/`Option`/`array`/`tuple` so a type's nested companions are emitted before the type that references them (self-contained AST regardless of declaration order).

`ffi/internal/c_wire.nim`: docstring only — the runtime still owns just the cstring helpers; composite allocations are inlined by the macro.

## Tests

`tests/unit/test_c_wire.nim` — three suites, 7 tests:

- **`Outer`/`Inner`**: nested struct, `seq[Inner]`, `seq[string]`, `seq[byte]`, `Option[string]`, `Option[int]`, `seq[Option[T]]`, plus edge cases (empty strings/seqs, `none`, `some("")` not collapsing to `none`).
- **`Shapes`**: `array` (POD, GC'd-element, nested-`{.ffi.}`), `array` of `array`, named `tuple` (POD and nested-`{.ffi.}`), `tuple` in `array`, `array` in `Option`, `tuple` in `seq` — fully-populated, all-empty/none, and GC'd-contents round-trips.

## CI

Adds the `c-wire` job running `test_c_wire` across the **full OS matrix** (the codec is malloc/free- and layout-sensitive, so Linux-only coverage isn't enough).

## Verification

- `test_c_wire`: 7/7 pass under `--mm:orc` and `--mm:refc`
- 7/7 clean under ASAN+UBSAN (`-d:useMalloc -fsanitize=address,undefined`) — no leaks or UAF on the new `seq`/`Option`/`array`/`tuple` alloc/free paths
- All changed Nim files nph-clean

## Notes

- `abiCodegenImplemented` stays unflipped — `c` is still gated on proc/ctor/dtor/event dispatch. A later PR flips that seam and wires the C++/Rust foreign generators; this PR only completes the type codec.
- Follow-up from review (compile-time ABI-version mangling so a signature change fails the build) is tracked separately and intentionally out of scope here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,6 +80,18 @@ jobs:
       nim-versions:   ${{ needs.versions.outputs.nim-versions }}
       nimble-version: ${{ needs.versions.outputs.nimble }}
 
+  c-wire:
+    # The `c`-ABI cwire codec is layout-/allocator-sensitive (malloc/free, flat
+    # struct packing, allocShared for seq/Option), so cover it across the full
+    # OS matrix, not just Linux.
+    name: C Wire Codec
+    needs: versions
+    uses: ./.github/workflows/test.yml
+    with:
+      test: test_c_wire
+      nim-versions:   ${{ needs.versions.outputs.nim-versions }}
+      nimble-version: ${{ needs.versions.outputs.nimble }}
+
   cpp-e2e:
     # Codegen output doesn't vary with mm, so we matrix over OS and Nim only.
     # Windows runs MSVC by default and may surface codegen tweaks needed in
--- a/ffi/internal/c_macro_helpers.nim
+++ b/ffi/internal/c_macro_helpers.nim
@@ -2,8 +2,10 @@
 ## For each `{.ffi: "abi = c".}` object T, emits a `T_CWire` companion plus
 ## `cwirePack` / `cwireUnpack` / `cwireFree`. Field mapping: `string`→`cstring`,
 ## `seq[T]`→`<name>_items`+`<name>_len`, `Option[T]`/`Maybe[T]`→`ptr T_w`
-## (nil=none), nested {.ffi.}→`T_CWire`, POD unchanged; seq/Option nest to any
-## depth, but a `seq` may not nest inside seq/Option (no single-field form).
+## (nil=none), nested {.ffi.}→`T_CWire`, `array[N, T]`→inline `array[N, T_w]`,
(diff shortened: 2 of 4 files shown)


Ivansete-status (approved):
Maybe pending to support `tuple[a: T, b: F]` for GC'ed types and `array[N, T]` for GC'ed types too.
Besides, is also relevant to be able to distinguish several ABI versions at compile time. For example, if a function changes one param type, the compiler should complain that the lib function changed. Maybe this can be done by applying a _mangling_-like mechanism.

Issue #90, opened 2026-06-24
Replace per-request global lock with an MPSC request transport


gmelodie:
## Problem

`sendRequestToFFIThread` (`ffi/ffi_thread.nim`) wraps the whole `trySend + reqSignal.fireSync + reqReceivedSignal.waitSync` cycle in `ctx.lock`, serializing every foreign-thread call on a single mutex. This is the dominant scalability ceiling. The lock is load-bearing because the accept handshake uses a *shared* `reqReceivedSignal` — all callers wait on the same `ThreadSignalPtr`.

`FFIContext[T].reqChannel` is `ChannelSPSCSingle[ptr FFIThreadRequest]` (single-producer, single-slot).

## Proposal

1. **MPSC ingress** — replace `ChannelSPSCSingle` with an MPSC channel (`taskpools/channels_mpsc_unbounded`, already a transitive dep, or a bounded lock-free ring if submit backpressure is wanted). Update `initContextResources`/`deinitContextResources`.
2. **Per-request completion** — move the accept-ack off the shared `reqReceivedSignal`: embed a completion primitive (`ThreadSignalPtr` or `Atomic[uint8]`) in `FFIThreadRequest`, or drop the synchronous accept entirely (the callback already carries success/error; with an unbounded queue, send can't fail). This is the load-bearing change — do **not** land the queue swap alone.
3. **Drain loop** — change `ffiThreadBody` recv to `while tryRecv: dispatch` and fire each request's own ack.
4. Delete the now-unused shared `reqReceivedSignal` (frees one fd per context against the `MaxFFIContexts * 2` budget).

## Gotchas

- Removing the lock must not reintroduce a shared-signal race.
- MPSC node alloc must stay off the per-thread ORC `MemRegion` (use `c_malloc`-backed or intrusive link nodes — same dangling-TLS hazard documented in `ffi_thread_request.nim`).
- Shutdown must still drain so each `deleteRequest` defer runs; keep `allFutures(pending)`.

## Tests

- Stress test (K threads × M submits): all callbacks fire exactly once, no leaks, under asan/ubsan + tsan.
- Keep `test_ffi_context.nim` green under orc + refc.
- `bench/` submit-throughput at 1/2/4/8 producer threads (the headline number).

Effort: medium-large (queue swap is small; correct per-request completion + the concurrent leak/sanitizer proof is the bulk).

PR #97, opened 2026-06-24
test(bench): submit-throughput scaling gate for sendRequestToFFIThread (issue #90)

gmelodie:
## Summary

Adds a concurrent-submit stress test / benchmark for `sendRequestToFFIThread` (`ffi/ffi_thread.nim`) that turns [issue #90](https://github.com/logos-messaging/nim-ffi/issues/90) into a CI gate: it is **red today and only goes green once the per-request global lock is replaced**. Test + CI only; no library code changes.

## Why

`sendRequestToFFIThread` wraps the whole `trySend + reqSignal.fireSync + reqReceivedSignal.waitSync` cycle in `ctx.lock`. The lock is load-bearing because `reqChannel` is single-slot (`ChannelSPSCSingle`) and the accept handshake waits on a *shared* `reqReceivedSignal`, so every foreign-thread submit serialises — adding producer threads buys nothing. Issue #90 proposes MPSC ingress + per-request completion so submits become parallel lock-free enqueues. This PR is the executable spec that forces and then guards that change.

## What the test measures

`tests/bench/bench_ffi_submit.nim` fans K producer threads (1 → 8) at one context, each firing the same per-thread volume of no-op requests, and times the **submit phase only** — from the start gate until every producer returns from its last `sendRequestToFFIThread`. Completion is bounded by the single FFI consumer thread and is deliberately excluded, because the submit path is what #90 parallelises. Each thread count runs `FFI_SUBMIT_ITERS` times (default 5) and the **median** submit/sec is reported, so run-to-run noise can't move the verdict.

It also doubles as a correctness stress test: the aggregate callback count must match the submit count exactly (no drops or double-fires), with zero submit errors and — under asan/lsan/tsan — zero leaks or races.

## The scaling gate

By default the bench fails (non-zero exit) unless 8-thread submit throughput is at least `1.5x` the 1-thread rate. It cannot pass while the lock serialises submits.

Baseline measured on a 16-core Linux host (orc, `-d:danger`, median of 5): submit scaling held at `0.98–1.16x` across threads — flat, as the lock dictates. `1.5x` sits above that noise ceiling (so the lock-bound code fails reliably) and well below the `>=2x` that parallel lock-free enqueue yields on any multicore host (so the #90 fix clears it with margin). The gate hard-fails on degenerate config (`FFI_SUBMIT_PER_THREAD`/`FFI_SUBMIT_ITERS < 1`) so it can't be silently neutered.

## CI

New non-sanitized `Submit Scaling Gate` job (`.github/workflows/ci.yml`) runs the gate on Linux/orc — this is the check that stays red until #90 lands. It runs through a new `nimble-job.yml` reusable workflow: single-OS toolchain + cached deps that then runs one nimble command across the Nim matrix, for jobs whose only difference is the command (the gate, and now `check-bindings`, which was migrated onto the same helper to drop its duplicated inline setup). Multi-OS / sanitizer matrices stay in `test.yml` / `tests-sanitized.yml`. The sanitized jobs run the same bench with `FFI_SCALING_GATE=0` for leak/race coverage only, since sanitizer instrumentation makes throughput scaling meaningless. The `bench_ffi_submit` nimble task honors `NIM_FFI_SAN` / `NIM_FFI_MM` (shared `mmModes` / `applyTsanSuppressions` helpers, also adopted by `test_sanitized`).

## Test plan

- Verified locally (Nim 2.2.6) under orc and refc, and under asan/ubsan/lsan (orc) and tsan (orc + refc): exactly-once aggregate, no submit errors, no leak/race reports.
- Gate ON fails reliably (median scaling 0.98–1.18x across repeats, never near 1.5x); gate OFF and degenerate-config paths behave as intended.

## Notes / limitations

The gate intentionally turns the PR red until #90 is implemented; if branch protection requires all checks green, this blocks merge by design (flip the job to `continue-on-error` if you'd rather it be advisory). The gate must run non-sanitized: under tsan/asan the scaling signal is invalid. "Exactly once" is verified in aggregate (count equals N*M), not per-request identity.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -169,51 +169,15 @@ jobs:
           nimble test_cpp_e2e -y
 
   check-bindings:
+    # Single OS is enough — codegen output is platform-independent; the Nim
+    # matrix catches version-sensitive output (the PR #39 drift class).
     name: Check generated bindings
     needs: versions
-    # Codegen output is platform-independent — single OS is enough. Matrix
-    # over Nim versions to catch any version-sensitive output. Catches the
-    # class of drift surfaced in PR #39 (C++ regen committed, Rust
-    # overlooked); see `nimble check_bindings` in ffi.nimble.
-    strategy:
-      fail-fast: false
-      matrix:
-        nim-version: ${{ fromJSON(needs.versions.outputs.nim-versions) }}
-    runs-on: ubuntu-22.04
-    env:
-      NIMBLE_VERSION: ${{ needs.versions.outputs.nimble }}
-    steps:
-      - uses: actions/checkout@v4
-
-      - name: Setup Nim
-        uses: jiro4989/setup-nim-action@v2
-        with:
-          nim-version: ${{ matrix.nim-version }}
-          repo-token: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Install Nimble ${{ env.NIMBLE_VERSION }}
-        run: |
-          cd /tmp && nimble install "nimble@${{ env.NIMBLE_VERSION }}" -y
-          echo "$HOME/.nimble/bin" >> $GITHUB_PATH
-
-      - name: Cache nimble deps
-        id: cache-nimbledeps
-        uses: actions/cache@v4
-        with:
-          path: |
-            nimbledeps/
-            nimble.paths
-          key: ${{ runner.os }}-nimbledeps-${{ matrix.nim-version }}-${{ hashFiles('*.nimble') }}
-          restore-keys: |
-            ${{ runner.os }}-nimbledeps-${{ matrix.nim-version }}-
-            ${{ runner.os }}-nimbledeps-
-
-      - name: Install nimble deps
-        if: steps.cache-nimbledeps.outputs.cache-hit != 'true'
-        run: nimble setup --localdeps -y
-
-      - name: Verify checked-in bindings match generator output
-        run: nimble check_bindings -y
+    uses: ./.github/workflows/nimble-job.yml
+    with:
+      run: nimble check_bindings -y
+      nim-versions:   ${{ needs.versions.outputs.nim-versions }}
+      nimble-version: ${{ needs.versions.outputs.nimble }}
 
   tests-asan-ubsan:
     name: Tests · ASan+UBSan+LSan
@@ -230,6 +194,19 @@ jobs:
     uses: ./.github/workflows/tests-sanitized.yml
     with:
       sanitizer:      tsan
+      nim-versions:   ${{ needs.versions.outputs.nim-versions }}
+      nimble-version: ${{ needs.versions.outputs.nimble }}
+
+  submit-scaling-gate:
+    # Intentionally RED until the per-request submit lock is replaced: asserts
+    # sendRequestToFFIThread submit throughput scales with producer threads,
+    # which the global lock cannot do. Non-sanitized + orc so the timing is
+    # meaningful (sanitizers distort thread scheduling). See the bench source.
+    name: Submit Scaling Gate
+    needs: versions
+    uses: ./.github/workflows/nimble-job.yml
+    with:
+      run: NIM_FFI_MM=orc FFI_SUBMIT_PER_THREAD=20000 nimble bench_ffi_submit -y
       nim-versions:   ${{ needs.versions.outputs.nim-versions }}
       nimble-version: ${{ needs.versions.outputs.nimble }}
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,10 +198,12 @@ jobs:
       nimble-version: ${{ needs.versions.outputs.nimble }}
 
   submit-scaling-gate:
-    # Intentionally RED until the per-request submit lock is replaced: asserts
-    # sendRequestToFFIThread submit throughput scales with producer threads,
-    # which the global lock cannot do. Non-sanitized + orc so the timing is
-    # meaningful (sanitizers distort thread scheduling). See the bench source.
+    # Forcing function, red by design: asserts sendRequestToFFIThread submit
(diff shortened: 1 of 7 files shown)


.github/workflows/nimble-job.yml
```diff
@@ -0,0 +1,66 @@
+name: nimble-job
+
+# Single-OS job that sets up the Nim/Nimble toolchain + cached deps, then runs
```
Ivansete-status, line 3: This file is great but seems unrelated to the PR's purpose?

.github/workflows/ci.yml
```diff
@@ -233,6 +197,19 @@ jobs:
...
+  submit-scaling-gate:
+    # Intentionally RED until the per-request submit lock is replaced: asserts
+    # sendRequestToFFIThread submit throughput scales with producer threads,
+    # which the global lock cannot do. Non-sanitized + orc so the timing is
+    # meaningful (sanitizers distort thread scheduling). See the bench source.
+    name: Submit Scaling Gate
+    needs: versions
+    uses: ./.github/workflows/nimble-job.yml
+    with:
+      run: NIM_FFI_MM=orc FFI_SUBMIT_PER_THREAD=20000 nimble bench_ffi_submit -y
```
Ivansete-status, line 211: What happens in refc?

.github/workflows/ci.yml
```diff
@@ -233,6 +197,19 @@ jobs:
       nim-versions:   ${{ needs.versions.outputs.nim-versions }}
       nimble-version: ${{ needs.versions.outputs.nimble }}
 
+  submit-scaling-gate:
+    # Intentionally RED until the per-request submit lock is replaced: asserts
+    # sendRequestToFFIThread submit throughput scales with producer threads,
+    # which the global lock cannot do. Non-sanitized + orc so the timing is
+    # meaningful (sanitizers distort thread scheduling). See the bench source.
```
Ivansete-status: I'd give an explanatory comment instead of a PR/temporary one.

tests/bench/bench_ffi_submit.nim
```diff
@@ -0,0 +1,173 @@
+## Concurrent-submit stress test + throughput bench for `sendRequestToFFIThread`,
+## motivating its per-request submit lock. See tests/bench/README.md for the why.
+
+import std/[atomics, algorithm, strutils, os]
+import results
+import ../../ffi # chronos (Moment/Duration) and the FFI surface both arrive here.
+
+type BenchLib = object
+
+registerReqFFI(NoopRequest, lib: ptr BenchLib):
```
Ivansete-status, line 10: Maybe, instead, we can use the `ctor` pragma for an arbitrary proc, but I guess it needs to use the explicit approach.

tests/bench/bench_ffi_submit.nim
```diff
@@ -0,0 +1,173 @@
...
+    retCode: cint, msg: ptr cchar, len: csize_t, userData: pointer
+) {.cdecl, gcsafe, raises: [].} =
+  let counter = cast[ptr Atomic[int]](userData)
+  discard counter[].fetchAdd(1)
+
+type ProducerArg = object
+  ctx: ptr FFIContext[BenchLib]
+  count: int
+
+proc producerBody(arg: ptr ProducerArg) {.thread, gcsafe.} =
```
Ivansete-status, line 34: ~~Should we add an assert in test that confirms the producer is actually producing and not discarding?~~

I saw it in enforceScalingGate

tests/bench/bench_ffi_submit.nim
```diff
@@ -0,0 +1,173 @@
...
+    formatFloat(RequiredScaling, ffDecimal, 2),
+    "x — submit path scales."
+
+proc main() =
+  let perThread = parseInt(getEnv("FFI_SUBMIT_PER_THREAD", "20000"))
+  let iters = parseInt(getEnv("FFI_SUBMIT_ITERS", "5"))
+  let gateOn = getEnv("FFI_SCALING_GATE", "1") != "0"
+  if perThread < 1 or iters < 1:
+    quit("FFI_SUBMIT_PER_THREAD and FFI_SUBMIT_ITERS must be >= 1")
+  let threadCounts = [1, 2, 4, 8]
```
Ivansete-status, line 122: What happens with 100?

ffi.nimble
```diff
@@ -104,23 +134,10 @@ task test_cpp_e2e, "Build and run the C++ end-to-end tests for the timer example
...
-      @[nimFlagsOrc, nimFlagsRefc]
   if san == "tsan":
-    let suppPath = thisDir() & "/tsan.supp"
-    let existing = getEnv("TSAN_OPTIONS")
-    if existing == "":
-      putEnv("TSAN_OPTIONS", "suppressions=" & suppPath)
-    elif "suppressions=" notin existing:
-      putEnv("TSAN_OPTIONS", existing & ":suppressions=" & suppPath)
-  for flags in modes:
+    applyTsanSuppressions()
```
Ivansete-status, line 139: Nice cleanup in this snippet. Thanks!

Ivansete-status (approved):
LGTM! Thanks for it! 🙌 
Just added some nits that I hope you find useful.

.github/workflows/ci.yml
gmelodie: The lock contention is memory-management-independent. Did you want me to add refc here as well?

tests/bench/bench_ffi_submit.nim
gmelodie: The point here is to understand how it behaves differently with increasing number of threads. So we could put 100 (why not 1000!) but that wouldn't change the conclusion since that's based on comparison between runs. In other words, if we were to put 100, it'd make more sense to have the following
```nim
let threadCounts = [100, 200, 400, 800]
```
but that would just oversubscribe the 2 cores on the CI, be slower and arrive at the same conclusion

PR #98, opened 2026-06-25
feat: lock-free MPSC request transport for the FFI thread

gmelodie:
## Summary

Replaces the single-slot SPSC channel + global submit lock that serialised every foreign-thread FFI call with a lock-free multi-producer / single-consumer queue (Vyukov's intrusive MPSC). Stacked on #97.

The old `sendRequestToFFIThread` took `ctx.lock`, `trySend`'d into a single-slot `ChannelSPSCSingle`, fired `reqSignal`, then **blocked on `reqReceivedSignal`** waiting for the FFI thread to ack each request one at a time. Every concurrent caller serialised behind that lock + ack handshake, even though completion is already reported asynchronously through each request's own callback.

### What changes

- **`ffi/ffi_request_queue.nim` (new):** `FFIRequestQueue`, an intrusive MPSC queue of `ptr FFIThreadRequest`. Producers enqueue with one wait-free `head.exchange`; the FFI thread is the sole consumer. The request *is* the queue node (`next` link reused intrusively), so enqueue allocates nothing. The stub sentinel is embedded inline in the queue (hence inline in `FFIContext`) — no separate heap node, so an intentionally-abandoned context leaks nothing extra and nothing lands on a per-thread ORC `MemRegion` (the dangling-TLS hazard already documented in `ffi_thread_request.nim`).
- **`ffi/ffi_thread_request.nim`:** adds the intrusive `next: Atomic[ptr FFIThreadRequest]` link to `FFIThreadRequest`, initialised in `allocBaseRequest`.
- **`ffi/ffi_thread.nim`:** `sendRequestToFFIThread` now just `push`es and fires the wake; the `timeout` param, the lock, the `trySend`, and the `reqReceivedSignal` wait are gone. A failed wake is non-fatal — the request is queued and the new `drainQueue()` dispatches it on the next poll tick, so we return `ok()` rather than double-firing the caller's error callback. `drainQueue()` drains to empty (fireSync coalesces, so one wake can cover many submits) and ticks the heartbeat per dispatch so a large backlog can't trip the wedged-FFI-thread detector mid-drain.
- **`ffi/ffi_context.nim`:** drops `lock`, `reqChannel`, and `reqReceivedSignal`; adds `reqQueue`. Init/deinit updated — the inline stub needs no teardown, just `initRequestQueue` rebinds head/tail on reuse.
- **`ffi/codegen/rust.nim` + `examples/timer/rust_bindings/src/api.rs`:** the generated `unsafe impl Send/Sync` SAFETY comment is rewritten to describe the lock-free enqueue rather than the old lock-serialisation argument.
- **`ffi/internal/ffi_macro.nim` + `tests/unit/test_ffi_context.nim`:** comment/doc references to `ctx.lock` updated to "the MPSC ingress hand-off"; the reentrancy-guard rationale updated (a re-dispatching handler would enqueue work its own blocked dispatcher can never drain).

### Notes

- The reentrancy guard is retained and still returns an `Err` immediately.
- No PR template in this repo.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -169,51 +169,15 @@ jobs:
           nimble test_cpp_e2e -y
 
   check-bindings:
+    # Single OS is enough — codegen output is platform-independent; the Nim
+    # matrix catches version-sensitive output (the PR #39 drift class).
     name: Check generated bindings
     needs: versions
-    # Codegen output is platform-independent — single OS is enough. Matrix
-    # over Nim versions to catch any version-sensitive output. Catches the
-    # class of drift surfaced in PR #39 (C++ regen committed, Rust
-    # overlooked); see `nimble check_bindings` in ffi.nimble.
-    strategy:
-      fail-fast: false
-      matrix:
-        nim-version: ${{ fromJSON(needs.versions.outputs.nim-versions) }}
-    runs-on: ubuntu-22.04
-    env:
-      NIMBLE_VERSION: ${{ needs.versions.outputs.nimble }}
-    steps:
-      - uses: actions/checkout@v4
-
-      - name: Setup Nim
-        uses: jiro4989/setup-nim-action@v2
-        with:
-          nim-version: ${{ matrix.nim-version }}
-          repo-token: ${{ secrets.GITHUB_TOKEN }}
-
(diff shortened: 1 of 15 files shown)


gmelodie:
closing in favor of #101

PR #99, opened 2026-06-25
feat(ffi): scalar type mappings + seq[byte] byte-string codec

gmelodie:
## Summary

The C++/CDDL codegen mishandled two type families used across `{.ffi.}` objects and procs, breaking the cross-language CBOR wire contract. This fixes both so the generated headers compile and round-trip byte-for-byte with the Nim `cbor_serialization` side.

### Problem 1 — missing scalar type mappings

`nimTypeToCpp` (`ffi/codegen/cpp.nim`) only mapped a subset of scalars; everything else passed through verbatim as a literal C++ name. So a `uint64` field emitted the non-type `uint64`, and `seq[byte]` emitted `std::vector<byte>` (collides with `std::byte`, no codec) — both fail to compile.

Extended the `case` to map the unsigned/odd-width scalars to fixed-width C++ types (`uint`/`uint64`→`uint64_t`, `uint32`/`uint16`/`uint8`/`byte`, `int16`/`int8`, `float32`→`float`). `seq[byte]` now resolves to `std::vector<uint8_t>`, at which point the existing `uint64_t` etc. CBOR overloads already in scope resolve cleanly.

`nimTypeToCddl` (`ffi/codegen/cddl.nim`) gains a `byte`→`uint` scalar case for completeness.

### Problem 2 — `seq[byte]` wire-format mismatch

Nim's `cbor_serialization` has a dedicated `seq[byte]` writer that emits a **CBOR byte string** (major type 2). The generated C++ side had no special case: `seq[byte]`→`std::vector<...>` hit the generic `std::vector<T>` codec, which emits/expects a **CBOR array** (major type 4). Result: decode errors in both directions.

- `cbor_helpers.hpp.tpl`: added concrete non-template `encode_cbor`/`decode_cbor` overloads for `std::vector<std::uint8_t>` using TinyCBOR's `cbor_encode_byte_string` / `cbor_value_is_byte_string` + `cbor_value_copy_byte_string`. Being non-template, they win overload resolution over the `std::vector<T>` template automatically — no codegen-site change needed.
- `cddl.nim`: `seq[byte]`/`seq[uint8]` now map to `bytes` in the schema, reflecting the actual wire shape.

## Tests

- `tests/unit/test_wire_compat.nim` — new `seq[byte]` suite pinning the exact golden hex (`a164626c6f6243010203`) and asserting the value's CBOR major type is a byte string (`0x40`–`0x5b`), never an array (`0x80`–`0x9b`). This is the cross-language contract the C++ side must match.
- `tests/e2e/cpp/test_timer_e2e.cpp` — `WireCompat` gtests encode a `std::vector<std::uint8_t>` and assert the bytes equal the Nim goldens (`43 01 02 03` and `40`), plus decode round-trip.
- Existing `test_serial.nim` / `test_c_wire.nim` `seq[byte]`/`blob` coverage confirmed still green.

## Verification

- `test_wire_compat`, `test_serial`, `test_cddl_codegen`, `test_meta` all pass under `--mm:orc`.
- Compiled+ran the C++ codec against vendored TinyCBOR (`g++ -std=c++17`): the bare codec, the exact codegen field-spelling (`std::vector<uint8_t> data;` struct), and a check that non-byte vectors still ride as CBOR arrays — all OK.

## Out of scope

Rust codegen (`ciborium`/`serde`) has the same mismatch (`Vec<u8>` serializes as a CBOR array without `#[serde(with = "serde_bytes")]`). cbind only generates cpp/cddl, so it doesn't block this — tracked as a follow-up for when Rust generation is enabled.

### DIFF
--- a/ffi/codegen/templates/cpp/cbor_helpers.hpp.tpl
+++ b/ffi/codegen/templates/cpp/cbor_helpers.hpp.tpl
@@ -36,6 +36,14 @@ inline CborError encode_cbor(CborEncoder& e, const std::vector<T>& v) {
     return cbor_encoder_close_container(&e, &arr);
 }
 
+// `seq[byte]` rides the wire as a CBOR byte string (major type 2), matching
+// Nim's cbor_serialization. This non-template overload beats the std::vector<T>
+// template in overload resolution, so std::vector<std::uint8_t> fields use it
+// automatically.
+inline CborError encode_cbor(CborEncoder& e, const std::vector<std::uint8_t>& v) {
+    return cbor_encode_byte_string(&e, v.data(), v.size());
+}
+
 template<typename T>
 inline CborError encode_cbor(CborEncoder& e, const std::optional<T>& v) {
     if (!v) return cbor_encode_null(&e);
@@ -111,6 +119,20 @@ inline CborError decode_cbor(CborValue& it, std::vector<T>& out) {
         if (err) return err;
     }
     return cbor_value_leave_container(&it, &inner);
+}
+
+// Counterpart to the byte-string encoder above: decode a CBOR byte string
+// (major type 2) back into std::vector<std::uint8_t>.
+inline CborError decode_cbor(CborValue& it, std::vector<std::uint8_t>& out) {
+    if (!cbor_value_is_byte_string(&it)) return CborErrorImproperValue;
+    size_t len = 0;
+    CborError err = cbor_value_get_string_length(&it, &len);
+    if (err) return err;
+    out.resize(len);
+    err = cbor_value_copy_byte_string(
+        &it, out.empty() ? nullptr : out.data(), &len, nullptr);
+    if (err) return err;
+    return cbor_value_advance(&it);
 }
 
 template<typename T>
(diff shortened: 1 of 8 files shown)


ffi/codegen/templates/cpp/cbor_helpers.hpp.tpl
```diff
@@ -113,6 +121,20 @@ inline CborError decode_cbor(CborValue& it, std::vector<T>& out) {
...
+inline CborError decode_cbor(CborValue& it, std::vector<std::uint8_t>& out) {
+    if (!cbor_value_is_byte_string(&it)) return CborErrorImproperValue;
+    size_t len = 0;
+    CborError err = cbor_value_get_string_length(&it, &len);
+    if (err) return err;
+    out.resize(len);
+    err = cbor_value_copy_byte_string(
+        &it, out.empty() ? nullptr : out.data(), &len, nullptr);
+    if (err) return err;
+    return cbor_value_advance(&it);
```
Ivansete-status: It seems these two lines should always go together. Therefore, is interesting to encapsulate them.

Ivansete-status (approved):

PR #101, opened 2026-06-25
feat(ffi): sharded MPSC request ingress (alternative to lock-free #98)

Ivansete-status:
## Summary

A simpler, faster alternative to the lock-free (Vyukov) request transport in #98, proposed so the two can be compared head-to-head against the #97 submit-scaling gate. Replaces the single-slot SPSC channel + global submit lock with a **sharded, mutex-guarded MPSC ingress**.

> Note: this branch builds on #97, so the diff vs `master` also carries #97's submit benchmark (`tests/bench/bench_ffi_submit.nim`, the `Submit Scaling Gate` CI job, and the nimble task). That is intentional — it makes the comparison runnable standalone.

## The problem (issue #90 / #97)

`sendRequestToFFIThread` held `ctx.lock` while **blocking** on `reqReceivedSignal` waiting for the FFI thread to ack each request, so concurrent producers could never overlap — the #97 gate stays red until that serialisation is removed.

## Why sharding, not lock-free

A **single** queue — lock-free or not — funnels every producer through **one shared cache line** (the lock, or the lock-free head), plus #98 keeps a `fireSync()` **syscall on every submit**. That hotspot caps aggregate throughput, so it can't scale past one thread. Measured on this hardware the lock-free single queue actually *anti-scales*.

This PR instead:
- **Shards the ingress into 16 independent per-producer lanes** (`ffi/ffi_request_queue.nim`) — removes the shared write point; producers contend only when two land on the same lane.
- **Coalesces the wake** — fires only on a lane's empty→non-empty edge, so submits don't each pay a syscall (the consumer drains to empty and re-polls every 100 ms, so a skipped edge can't strand a request).
- Keeps each lane a **plain intrusive FIFO under its own `Lock`**: trivially data-race-free under TSAN, no memory-ordering reasoning, no transient-nil/shutdown-race. The request doubles as its own queue node (intrusive `next`), so enqueue allocates nothing and never touches a Nim GC heap (the cross-thread `MemRegion` hazard in `ffi_thread_request.nim`).

## Benchmark comparison

Apple M5 (10 cores), orc, `-d:danger`, median of 5 — submit/sec and scaling vs 1 thread:

| threads | **sharded (this PR)** | vs 1T | lock-free MPSC (#98) | vs 1T |
| ---: | ---: | :---: | ---: | :---: |
| 1   | 2.14M | 1.00x | 1.18M | 1.00x |
| 8   | 17.6M | 8.22x | 0.23M | 0.20x |
| 16  | 20.3M | 9.52x | 0.13M | 0.11x |
| 32  | 23.4M | 10.96x | 0.115M | 0.10x |
| 64  | 24.5M | 11.44x | 0.113M | 0.10x |
| 100 | 23.8M | **11.13x** | 0.113M | **0.10x** |

The sharded ingress tracks the hardware ceiling (~24M/s plateau) and degrades gracefully past it; the single-hotspot lock-free queue collapses to a contention floor and gets *worse* with more threads (~211x slower at 100 threads). Both are correct at every level (callback count matches submits exactly, no drops/dupes).

This matters for the real workload: a foreign Go app can drive ~75 concurrent threads at the library — squarely in the regime where the single-hotspot design collapses and sharding holds.

## Trade-offs (drawbacks of sharding)

- **Per-lane FIFO, not global** — concurrent foreign callers already raced with no cross-thread ordering guarantee, so this matches the existing contract; a single thread's own order is preserved (it maps to one lane).
- **Lane count (16) is a fixed constant** — the knob that bounds how far it scales. ≥ expected concurrent producers keeps lane collisions near zero (75 Go threads → ~5/lane, still scales as the 100-thread row shows). On a much larger host it may warrant bumping; candidate follow-up: make it a `declareLibrary` setting.
- **Consumer takes O(lanes) locks per drain pass** (16 here) even when most lanes are empty — trivial at 16, grows if lanes are raised.
- Helps **many** producers, not one hot producer (single-lane-bound).
- Padding (~280 B/lane) assumes a 128-byte cache line.

## Compatibility

- **FFI context pool unchanged and functional** — `ffi_context_pool.nim` only drives the lifecycle procs; the benchmark exercises it (create/destroy per iteration) at up to 100 threads.
- Rust codegen + the generated `api.rs` SAFETY comment updated to describe the mutex-guarded enqueue.

## Test plan

- Unit suites pass under **orc + refc**: `test_ffi_context` (19), `test_event_thread`, `test_event_dispatch`, `test_gc_compat`.
- Library builds `--app:lib`.
- Submit gate passes with wide margin (11x) across 1→100 threads; exactly-once correctness holds.
- TODO before merge: confirm on the Linux CI host (should win there too — it removes contention regardless of OS), and run under asan/tsan via the sanitized job.

### DIFF
--- a/ffi/ffi_request_queue.nim
+++ b/ffi/ffi_request_queue.nim
@@ -41,7 +41,7 @@ static:
     "RequestQueueCount must be a power of two"
 
 type
-  RequestShard = object
+  RequestQueue = object
     lock: Lock
     head: ptr FFIThreadRequest ## consumer pops here (oldest)
     tail: ptr FFIThreadRequest ## producers on this queue append here (newest)
@@ -49,7 +49,7 @@ type
     pad: array[QueuePadBytes, byte]
 
   FFIRequestQueue* = object
-    queues: array[RequestQueueCount, RequestShard]
+    queues: array[RequestQueueCount, RequestQueue]
 
 var gRequestQueue {.threadvar.}: int
 var gRequestQueueAssigned {.threadvar.}: bool
@@ -64,26 +64,26 @@ proc myQueueIndex(): int {.raises: [].} =
   return gRequestQueue and (RequestQueueCount - 1) # RequestQueueCount is a power of two
 
 proc initRequestQueue*(q: var FFIRequestQueue) {.raises: [].} =
-  for i in 0 ..< RequestQueueCount:
-    q.queues[i].lock.initLock()
-    q.queues[i].head = nil
-    q.queues[i].tail = nil
-    q.queues[i].count = 0
+  for queue in q.queues.mitems:
+    queue.lock.initLock()
+    queue.head = nil
+    queue.tail = nil
+    queue.count = 0
 
 proc deinitRequestQueue*(q: var FFIRequestQueue) {.raises: [].} =
   ## Both producers and the consumer must have stopped. Frees any request still
   ## queued on any queue — e.g. one a producer raced in after the FFI thread's
   ## final drain — so a teardown race leaks nothing instead of dangling them.
-  for i in 0 ..< RequestQueueCount:
-    var node = q.queues[i].head
+  for queue in q.queues.mitems:
+    var node = queue.head
     while not node.isNil:
       let nxt = node[].next
       deleteRequest(node)
       node = nxt
-    q.queues[i].head = nil
-    q.queues[i].tail = nil
-    q.queues[i].count = 0
-    q.queues[i].lock.deinitLock()
+    queue.head = nil
+    queue.tail = nil
+    queue.count = 0
+    queue.lock.deinitLock()
 
 proc pushRequest*(
     q: var FFIRequestQueue, node: ptr FFIThreadRequest
@@ -113,23 +113,23 @@ proc detachAllRequests*(q: var FFIRequestQueue): ptr FFIThreadRequest {.raises: 
   ## (dispatch frees the node).
   var head: ptr FFIThreadRequest = nil
   var tail: ptr FFIThreadRequest = nil
-  for i in 0 ..< RequestQueueCount:
-    withLock q.queues[i].lock:
-      let h = q.queues[i].head
+  for queue in q.queues.mitems:
+    withLock queue.lock:
+      let h = queue.head
       if not h.isNil:
         if head.isNil:
           head = h
         else:
           tail[].next = h
-        tail = q.queues[i].tail
-        q.queues[i].head = nil
-        q.queues[i].tail = nil
-        q.queues[i].count = 0
+        tail = queue.tail
+        queue.head = nil
+        queue.tail = nil
+        queue.count = 0
   return head
 
 proc requestQueueLen*(q: var FFIRequestQueue): int {.raises: [].} =
   var n = 0
-  for i in 0 ..< RequestQueueCount:
-    withLock q.queues[i].lock:
-      n += q.queues[i].count
+  for queue in q.queues.mitems:
+    withLock queue.lock:
+      n += queue.count
   return n
--- a/ffi/ffi_request_queue.nim
+++ b/ffi/ffi_request_queue.nim
@@ -45,7 +45,6 @@ type
     lock: Lock
     head: ptr FFIThreadRequest ## consumer pops here (oldest)
     tail: ptr FFIThreadRequest ## producers on this queue append here (newest)
-    count: int ## queue depth, for metrics only
     pad: array[QueuePadBytes, byte]
 
   FFIRequestQueue* = object
@@ -68,7 +67,6 @@ proc initRequestQueue*(q: var FFIRequestQueue) {.raises: [].} =
     queue.lock.initLock()
     queue.head = nil
     queue.tail = nil
-    queue.count = 0
 
 proc deinitRequestQueue*(q: var FFIRequestQueue) {.raises: [].} =
   ## Both producers and the consumer must have stopped. Frees any request still
@@ -82,7 +80,6 @@ proc deinitRequestQueue*(q: var FFIRequestQueue) {.raises: [].} =
       request = nextRequest
     queue.head = nil
     queue.tail = nil
-    queue.count = 0
     queue.lock.deinitLock()
 
 proc pushRequest*(
@@ -103,7 +100,6 @@ proc pushRequest*(
     else:
       q.queues[idx].tail[].next = request
     q.queues[idx].tail = request
-    q.queues[idx].count.inc()
     return wasEmpty
 
 proc mergeQueues*(q: var FFIRequestQueue): ptr FFIThreadRequest {.raises: [].} =
@@ -125,12 +121,4 @@ proc mergeQueues*(q: var FFIRequestQueue): ptr FFIThreadRequest {.raises: [].} =
         tail = queue.tail
         queue.head = nil
         queue.tail = nil
-        queue.count = 0
   return head
-
-proc requestQueueLen*(q: var FFIRequestQueue): int {.raises: [].} =
-  var n = 0
-  for queue in q.queues.mitems:
-    withLock queue.lock:
-      n += queue.count
-  return n
--- a/ffi/ffi_request_queue.nim
+++ b/ffi/ffi_request_queue.nim
@@ -106,12 +106,12 @@ proc pushRequest*(
     q.queues[idx].count.inc()
     return wasEmpty
 
-proc detachAllRequests*(q: var FFIRequestQueue): ptr FFIThreadRequest {.raises: [].} =
-  ## Single-consumer: splice every queue's queued chain (each queue FIFO, linked
-  ## by `next`) into one chain and reset the queues to empty, taking each queue's
-  ## lock once. Returns nil when all queues are empty; the caller then owns every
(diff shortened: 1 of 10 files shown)


fcecin (approved):
LGTM!

ffi/ffi_request_queue.nim
```diff
@@ -0,0 +1,137 @@
...
+    ## lane collisions (hence lock contention) near zero.
+  LanePadBytes = 192
+    ## Pads each lane well past a cache line (128B on Apple silicon) so adjacent
+    ## lanes' hot fields never false-share — false sharing would re-serialise
+    ## exactly what the sharding is meant to spread out.
+
+static:
+  # `myLaneIndex` maps threads to lanes with an `and` mask, so the count must be
+  # a power of two — otherwise the distribution silently skews onto a subset.
+  doAssert (RequestLaneCount and (RequestLaneCount - 1)) == 0,
```
gmelodie: What is `lane`? how does it differ from `queue`? If not, can we not introduce another term?

gmelodie: Ah i see, so a queue is made up of several lanes? why not just 
`lane == queue`
`queue == queues`

ffi/ffi_request_queue.nim
```diff
@@ -0,0 +1,137 @@
...
+var gRequestLane {.threadvar.}: int
+var gRequestLaneAssigned {.threadvar.}: bool
+var gRequestLaneCounter: Atomic[int]
+  ## Hands each producer thread a distinct lane round-robin on first use, so
+  ## lanes fill evenly regardless of OS thread-id distribution.
+
+proc myLaneIndex(): int {.raises: [].} =
+  if not gRequestLaneAssigned:
+    gRequestLane = gRequestLaneCounter.fetchAdd(1)
+    gRequestLaneAssigned = true
```
gmelodie: Could this cause a race where two workers would assign at the same time? should `requestLaneAssigned` be atomic as well?

ffi/ffi_request_queue.nim
```diff
@@ -0,0 +1,137 @@
...
+  for i in 0 ..< RequestLaneCount:
+    var node = q.lanes[i].head
+    while not node.isNil:
+      let nxt = node[].next
+      deleteRequest(node)
+      node = nxt
+    q.lanes[i].head = nil
+    q.lanes[i].tail = nil
+    q.lanes[i].count = 0
+    q.lanes[i].lock.deinitLock()
```
gmelodie: Why not `for lane in q.lanes`?

ffi/ffi_request_queue.nim
```diff
@@ -0,0 +1,137 @@
...
+
+proc pushRequest*(
+    q: var FFIRequestQueue, node: ptr FFIThreadRequest
+): bool {.raises: [].} =
+  ## Append `node` to this producer thread's lane (one O(1) critical section).
+  ## The queue takes ownership. Returns true iff that lane was empty before the
```
gmelodie:
```suggestion
  ## The queue takes ownership. Returns true if that lane was empty before the
```

ffi/ffi_request_queue.nim
```diff
@@ -0,0 +1,137 @@
...
+  ## push — i.e. the caller should wake the consumer. When the lane was
+  ## non-empty the consumer is already draining (or has a wake pending), so
+  ## firing again is a wasted syscall; coalescing the wake is what lets
+  ## concurrent submits scale. A skipped edge is bounded by the consumer's 100ms
+  ## poll, so it can never strand a request.
```
gmelodie:
```suggestion
  ## non-empty the consumer is already draining (or has a wake pending), so
  ## firing again is a wasted syscall.
```

ffi/ffi_request_queue.nim
```diff
@@ -0,0 +1,137 @@
...
+  node[].next = nil
+  let idx = myLaneIndex()
+  withLock q.lanes[idx].lock:
+    let wasEmpty = q.lanes[idx].tail.isNil
+    if q.lanes[idx].tail.isNil:
```
gmelodie:
```suggestion
    if wasEmpty:
```

ffi/ffi_request_queue.nim
```diff
@@ -0,0 +1,137 @@
...
+    q.lanes[idx].tail = node
+    q.lanes[idx].count.inc()
+    return wasEmpty
+
+proc detachAllRequests*(q: var FFIRequestQueue): ptr FFIThreadRequest {.raises: [].} =
+  ## Single-consumer: splice every lane's queued chain (each lane FIFO, linked
+  ## by `next`) into one chain and reset the lanes to empty, taking each lane's
+  ## lock once. Returns nil when all lanes are empty; the caller then owns every
+  ## request in the chain and must read each node's `next` before dispatching it
+  ## (dispatch frees the node).
```
gmelodie:
```suggestion
proc mergeLanes*(q: var FFIRequestQueue): ptr FFIThreadRequest {.raises: [].} =
  ## Single-consumer: splice every lane into one chain and reset the lanes to empty.
  ## Returns nil when all lanes are empty;
  ## the caller then owns every request in the chain
  ## and must read each node's `next` before dispatching it.
```

ffi/ffi_request_queue.nim
```diff
@@ -0,0 +1,137 @@
...
+        q.lanes[i].tail = nil
+        q.lanes[i].count = 0
+  return head
+
+proc requestQueueLen*(q: var FFIRequestQueue): int {.raises: [].} =
+  var n = 0
+  for i in 0 ..< RequestLaneCount:
+    withLock q.lanes[i].lock:
+      n += q.lanes[i].count
+  return n
```
gmelodie: If we made count atomic we could get rid of this lock

ffi/ffi_thread.nim
```diff
@@ -122,30 +121,40 @@ proc ffiThreadBody[T](ctx: ptr FFIContext[T]) {.thread.} =
...
+        var node = ctx.reqQueue.detachAllRequests()
+        if node.isNil:
+          break
+        while not node.isNil:
+          let nxt = node[].next # read before processRequest frees the node
+          # Tick per dispatch so a large backlog can't flatline the heartbeat
+          # and trip the event thread's wedged-FFI-thread detection mid-drain.
+          discard ctx.ffiHeartbeat.fetchAdd(1)
+          if ctx.myLib.isNil():
+            ctx.myLib = addr ffiReqHandler
```
gmelodie, line 141: Can we move this out of this loop?

gmelodie (approved):
Couple nit-ish comments. Also, can you reduce a bit the walls of comments? It's a bit distracting

ffi/ffi_request_queue.nim
Ivansete-status: There is no data race because any consumer thread will get its own copy of each threadvar var.
The ~~gRequestLaneCounter~~ gRequestQueueCounter is atomic because it acts as a simple round robin load balancer, and it is global to all consumer threads.

ffi/ffi_request_queue.nim
Ivansete-status: I thought the same when I read it first but I think the `wasEmpty` gives a timing semantics that only make sense in the return statement.

ffi/ffi_request_queue.nim
Ivansete-status: Addressed in 83eeec7b7e46417f086884bebcc390dd1e02ae2a

ffi/ffi_request_queue.nim
Ivansete-status: Addressed in f4649999ccb06b0b54040f0a332e0b2909317678

ffi/ffi_request_queue.nim
Ivansete-status: That stands for "if and only if" ( kind of conditional math abbreviation ) that Claude loves to use :)
Addressed in 6a41c2ade77e5bc753140307304312fda2d601e5

ffi/ffi_request_queue.nim
Ivansete-status: Addressed in 8a826757af31e749899e5c859bf4f75a5c802ee9

ffi/ffi_request_queue.nim
Ivansete-status: I deleted this proc in 96375c2e5d9af57b0dbbc87a264edcd81d8dbbe3

ffi/ffi_thread.nim
Ivansete-status: Addressed in 9f46edaaafd7ac8105217d6af8f3f124a436774f

Ivansete-status: I reverted it in: 51e499445eff831c8b150f59f6c05fa090409375

ffi/ffi_thread.nim
```diff
@@ -88,6 +87,12 @@ proc ffiNotifyEventEnqueuedHook() {.gcsafe, raises: [].} =
     if res.isErr():
       error "failed to fire eventQueueSignal after enqueue", err = res.error
 
+proc proveAlive(ctx: ptr FFIContext) =
```
gmelodie, line 85: beautiful!

gmelodie (approved):

PR #102, opened 2026-06-29
feat(codegen): C binding generator (-d:targetLang=c)

gmelodie:
## Summary

Adds a C backend to the binding generator, alongside the existing Rust / C++ / CDDL targets, selected with `-d:targetLang=c`. It emits a header-only C binding split into three headers plus a `CMakeLists.txt`, speaking the same CBOR wire format as the other backends so a C consumer talks to the unmodified Nim dylib.

The motivation is reach: the C++ backend already proves the CBOR round-trip works for a header-only consumer, but it leans on templates, overloading, `std::vector`/`std::optional`/`std::string`, and RAII. Plenty of consumers (embedded, other-language FFI shims, build systems that won't take a C++ toolchain) want a plain C surface. This is that surface.

## Header layout

The binding is emitted as three headers so the example reads top-down instead of as one 1600-line wall:

- `nim_ffi_prelude.h` — owned string/byte types (`NimFfiStr` / `NimFfiBytes`) plus the libc / TinyCBOR includes every binding needs. Library-agnostic.
- `nim_ffi_cbor.h` — leaf CBOR codecs and the buffer drivers; `#include`s the prelude. Library-agnostic.
- `<lib>.h` — the library-specific structs, monomorphised codecs and the async API; `#include`s the cbor header.

Each file is independently include-guarded, so two nim-ffi bindings can share a translation unit.

## How it works

C has neither generics nor overloading, so the codecs the C++ backend gets for free from templates are **monomorphised** at generation time:

- Each distinct `seq[T]` / `Option[T]` becomes its own struct plus an `encode`/`decode`/`free` triple, emitted in dependency order via a small type registry (`ensureCType`).
- Leaf types (scalars, `string`→`NimFfiStr`, `seq[byte]`→`NimFfiBytes`) get distinctly-named codecs from the shared `nim_ffi_cbor.h` template.
- Typed `{.ffiEvent.}` listeners are emitted as function-pointer + `void*` user-data handlers with per-event trampolines.

## Async, binding-owned API (no sync wrappers, no manual free)

The high-level `<lib>_ctx_*` API is **asynchronous and callback-driven**. Each method and the constructor take a typed result callback and return immediately; the callback fires exactly once — synchronously on a submit-time failure, otherwise from the Nim dispatch thread when the reply lands. There is no blocking wait and no per-call timeout.

The binding **owns and reclaims all reply memory**, so the caller never frees anything:

- Request-side strings/sequences are borrowed (`nimffi_str(...)`); the binding only reads them while encoding.
- Reply values and error strings handed to a result callback are owned by the binding and valid only for the duration of that callback — copy out whatever you need to keep before returning. The generated `<lib>_free_<Type>()` helpers are internal (the trampolines use them to reclaim decoded payloads).
- The one handle the caller owns is the context delivered to the constructor callback, released with `<lib>_ctx_destroy()` — object lifecycle, not returned data.

This replaces the earlier blocking design (out-parameters, a heap `char** err` channel, and a pthread-backed reference-counted call helper), which is removed.

## Reuse / minimal duplication

- The vendored TinyCBOR copy under `ffi/codegen/templates/cpp/vendor` is shared by both backends — no second vendor.
- Codegen helpers common to C and C++ (per-proc envelope naming, lib-prefix stripping, proc classification, context-type derivation) were extracted into `ffi/codegen/c_cpp_common.nim`; `cpp.nim` now uses them with byte-identical output.
- The CMake logic that builds the Nim dylib + TinyCBOR was extracted into `ffi/codegen/templates/nim_ffi_lib.cmake`, `include()`d by both the C and C++ generated `CMakeLists.txt` (~90 lines removed from each C++ CMakeLists).

## Validation

- New `tests/unit/test_c_codegen.nim` drives `generateCLibHeader` (plus the `generateCPreludeHeader` / `generateCCborHeader` shared-header generators) against a synthetic registry and asserts on the emitted text, including the async callback signatures, the per-method trampolines, the async constructor, the three-way header split with its include guards, and the absence of any blocking-helper or `timeout_ms` machinery.
- New `tests/e2e/c` ctest harness builds the generated timer bindings against the real Nim dylib and asserts on the constructor, methods, nested `seq`/`Option` payloads, multi-parameter requests, the error channel, and the typed event listener.
- Generated headers compile clean under `-std=c11 -Wall -Wextra` (and when included from C++); the e2e runs green under ASan/UBSan/LSan, confirming the binding-owned model leaks nothing.
- New `nimble` tasks: `genbindings_c`, `genbindings_c_echo`, `check_bindings_c` (wired into `check_bindings`), `test_c_e2e`, `test_c_e2e_sanitized`.
- Checked-in C bindings are regenerated and gated for drift by `check_bindings_c`, which now diffs all three headers (`<lib>.h`, `nim_ffi_prelude.h`, `nim_ffi_cbor.h`) for both the timer and echo examples. The C++ timer output changes only in the intended CMake dedup; the echo `.hpp` is also regenerated (it had drifted from the shared `advance_if_ok` helper because `check_bindings_cpp` only diffed the timer binding — the gate now regenerates and diffs the echo binding too, matching `check_bindings_c`).


examples/echo/c_bindings/echo.h
```diff
@@ -0,0 +1,846 @@
...
+/* Generated C binding for a nim-ffi library. Requests/responses travel as
+ * CBOR (encoded with vendored TinyCBOR on this side, matching the Nim-side
+ * cbor_serial codec on the wire — both ends speak RFC 8949).
+ *
+ * Memory ownership contract:
+ *   - Request-side strings/sequences are *borrowed*: the binding only reads
+ *     them while encoding, so a string literal wrapped with nimffi_str() is
+ *     fine and is never freed by the binding.
+ *   - Response-side values returned through an out-parameter are *owned* by
+ *     the caller. Release them with the generated <lib>_free_<Type>() helper.
```
Ivansete-status: Having to use an external _free function is very memory leak prone. I'd rather use callbacks with a well defined lifetime, and avoid the lib consumer to manually call free on every API call.

Ivansete-status (commented):
Thanks for it!
Adding one comment so far.
I wonder if, besides cbor, we should also auto-generate for ABI=c? Or maybe is redundant in this particular case.

gmelodie:
>  I wonder if, besides cbor, we should also auto-generate for ABI=c? Or maybe is redundant in this particular case.

You mean the examples? I think we should not in this case, following convention from other examples. Either that or we generate abi=rust and abi=cpp for rust and cpp respectively as well.

examples/echo/c_bindings/echo.h
gmelodie: This only happens when using the sync convenience wrappers, no? callbacks is how it normally works, but user can choose the sync wrappers if they want to do `a = myFunc()`, however this comes with the drawbacks of them having to free the types themselves.

examples/timer/c_bindings/main.c
```diff
@@ -0,0 +1,209 @@
...
+ * the caller never frees them; it copies out whatever it wants to keep. This
+ * demo turns each async call back into a sequential step by polling a `done`
+ * flag (the same pattern the typed event listener already uses). */
+
+/* Poll up to ~5s for a callback to fire. Returns false if it never did, so the
+ * caller can report a stuck call instead of treating it as an empty success. */
+static bool wait_done(volatile int* done) {
+    for (int i = 0; i < 500 && !*done; i++) {
+        struct timespec t = {0, 10 * 1000 * 1000}; /* 10ms */
+        nanosleep(&t, NULL);
```
Ivansete-status: We may need to wrap that `nanosleep` call, and any other posix-only function, so that the example is portable to Windows too.

examples/echo/c_bindings/echo.h
```diff
@@ -0,0 +1,817 @@
+#ifndef NIM_FFI_PRELUDE_H_INCLUDED
+#define NIM_FFI_PRELUDE_H_INCLUDED
+/* Generated C binding for a nim-ffi library. Requests/responses travel as
+ * CBOR (encoded with vendored TinyCBOR on this side, matching the Nim-side
```
Ivansete-status: It sounds interesting to also emit an example with pure C ABI and without CBOR.
( and have echo_c_abi.h echo_cbor_abi.h, or any other name, to compare later on .)

examples/echo/c_bindings/echo.h
```diff
@@ -0,0 +1,817 @@
...
+    v.len = s ? strlen(s) : 0;
+    return v;
+}
+
+static inline void nimffi_free_str(NimFfiStr* v) {
+    if (!v || !v->data) {
+        return;
+    }
+    free(v->data);
+    v->data = NULL;
```
Ivansete-status: Should v = NULL ; too?

examples/echo/c_bindings/echo.h
```diff
@@ -0,0 +1,817 @@
...
+    *out = (uint32_t)tmp;
+    return CborNoError;
+}
+static inline CborError nimffi_dec_u16(CborValue* it, uint16_t* out) {
+    uint64_t tmp = 0;
+    CborError err = nimffi_dec_u64(it, &tmp);
+    if (err) {
+        return err;
+    }
+    *out = (uint16_t)tmp;
```
Ivansete-status: We need to have slightly better truncation protection when casting to a smaller type.

Something like:
    if (tmp < INT16_MIN || tmp > INT16_MAX) {
        return CborErrorDataTooLarge;
    }

examples/echo/c_bindings/echo.h
```diff
@@ -0,0 +1,817 @@
...
+#endif
+
+/* Result delivery callback exported by the Nim dylib: `ret` is 0 on success
+ * (then `msg`/`len` carry the CBOR response) or non-zero on failure (then
+ * `msg`/`len` carry the error text, which is NOT NUL-terminated). */
+typedef void (*FFICallback)(int ret, const char* msg, size_t len, void* user_data);
+
+/* RET_MISSING_CALLBACK from the Nim dispatcher: the callback will never fire,
+ * so the request path must report the failure itself. */
+#define NIMFFI_RET_MISSING_CALLBACK 2
```
Ivansete-status: I miss seeing explicit OK ( == 0 ) value.

examples/echo/c_bindings/echo.h
```diff
@@ -0,0 +1,817 @@
+#ifndef NIM_FFI_PRELUDE_H_INCLUDED
```
Ivansete-status: Given that this is an example and we don't need something that is easy to import, maybe we can split this into separated headers, each with clear purpose. That can help to understand the example.
Maybe:
- nim_ffi_prelude.h
- nim_ffi_cbor.h
- echo.h

examples/timer/c_bindings/main.c
```diff
@@ -0,0 +1,209 @@
...
+    }
+    return *done != 0;
+}
+
+static volatile int g_echo_count = 0;
```
Ivansete-status: A little bit of premature opt suggestion:
To make the atomicity more portable, we can use atomic header, at the top:
```nim
#if defined(__STDC_NO_ATOMICS__)
#  error "C11 atomics required (or provide a mutex/condvar fallback)"
#else
#  include <stdatomic.h>
#endif
```

examples/timer/c_bindings/main.c
```diff
@@ -0,0 +1,209 @@
...
+           w.text_a, w.num_a, w.num_b);
+
+    uint64_t handle =
+        my_timer_ctx_add_on_echo_fired_listener(ctx, on_echo_fired, NULL);
+    EchoRequest evt_req = {nimffi_str("event-demo"), 1};
+    memset(&w, 0, sizeof(w));
+    my_timer_ctx_echo(ctx, &evt_req, on_echo, &w);
+    wait_done(&w.done);
+    /* The event fires from the library's dispatch thread; give it a moment. */
+    struct timespec settle = {0, 100 * 1000 * 1000};
```
Ivansete-status: To give some more time
```suggestion
    struct timespec settle = {0, 500 * 1000 * 1000};
```

ffi/codegen/c_cpp_common.nim
```diff
@@ -0,0 +1,62 @@
+## Helpers shared by the language-specific binding generators (cpp.nim, c.nim).
```
Ivansete-status, line 1: I'd rename that module to c_cpp_common.nim

Ivansete-status (approved):
LGTM! Thanks for it! 🙌 
The only point I'm missing is seeing the c gen to work with pure c abi as well. If that makes sense and is reasonable.

examples/echo/c_bindings/echo.h
gmelodie: Created an issue https://github.com/logos-messaging/nim-ffi/issues/105 since this is true for all examples, not only C

PR #104, opened 2026-06-29
refactor(codegen): shared FFIType IR + single type parser for C/C++/Rust

gmelodie:
## Summary

The three FFI codegen backends each carried their own ad-hoc parser for the Nim type strings in the metadata registry (`FFIFieldMeta.typeName`, e.g. `"seq[Option[int]]"`): C recursed via `genericInnerType` + `startsWith("seq[")`, C++ via `genericInnerType`, and Rust via hand-rolled `t[4 .. ^2]` slicing. Three copies of the same leaf-scalar table, three places to fix any new shape — and they had already drifted: `nimTypeToRust` was missing the `int8/int16/uint8/uint16/uint32/byte/float32` cases that C and C++ handle, so those types silently fell through to `capitalizeFirstLetter(t)` and emitted invalid Rust.

This introduces one structured type model and one parser that all three backends consume.

- New `ffi/codegen/types_ir.nim`: a recursive `FFIType` descriptor (`ScalarKind` + `FFITypeKind`), the single `parseFFIType` parser, and a generic `renderNative` walker driven by a small per-language `NativeTypeMap`. `genericInnerType` moves here as a private parser helper (removed from `c_cpp_common.nim`).
- `cpp.nim` / `rust.nim`: `nimTypeToCpp` / `nimTypeToRust` are now `renderNative(map, parseFFIType(t))`. Public signatures unchanged, so callers are untouched. This fixes the Rust scalar drift by construction (the scalar set now lives in exactly one place).
- `c.nim`: `ensureCType` switches on `FFIType.kind` instead of string prefixes; the leaf scalar table is expressed once as `scalarCInfoTable` keyed by `ScalarKind`, and `leafSuffix`/`cToken` derive from it. Monomorphisation and codec-body emission (`emitSeqType`/`emitOptType`/`emitStructType`) are unchanged — only their input becomes the shared IR.

Codec-body emission and the high-level ctor/method/event wrappers are deliberately left per-backend (they genuinely diverge: `Result<T>` vs `std::future` vs flume, `unique_ptr` vs RAII vs `Drop`); a later phase could revisit them behind a backend vtable. `ftHandle` is parsed for completeness but no production caller passes a `handleNames` set (handles are pre-converted to `pointer` upstream), so that arm is currently exercised only by the unit tests.

## Verification

- New `tests/unit/test_rust_codegen.nim` pins the drift fix directly: `nimTypeToRust` over the full scalar set (including the previously-broken `int8`/`int16`/`uint8`/`uint16`/`uint32`/`byte`/`float32`), `seq[byte]` → `Vec<u8>`, `ptr T`/`pointer`, nested generics, `Maybe` == `Option`, and struct capitalisation.
- New `tests/unit/test_types_ir.nim` covers `parseFFIType` directly: the full scalar set, `seq[byte]`/`seq[uint8]` → `ftBytes`, nested `seq[Option[seq[int]]]`, `Maybe[T]` == `Option[T]`, handle names → `ftHandle`, `ptr T`/`pointer` → `ftPtr`, unknown name → `ftStruct`, plus `renderNative` round-trips.
- Golden regen + diff: regenerated the committed example bindings; the C and C++ outputs for `examples/echo` and `examples/timer` are byte-identical to what is on disk, and the Rust output is identical (the examples don't use any of the previously-broken scalar types, so the drift fix changes no checked-in output).
- All codegen unit tests pass under both `--mm:orc` and `--mm:refc` (`test_types_ir`, `test_rust_codegen`, `test_c_codegen`, `test_cddl_codegen`, `test_string_helpers`, `test_meta`).

### DIFF
--- a/ffi/codegen/c.nim
+++ b/ffi/codegen/c.nim
@@ -13,7 +13,7 @@
 ## distinctly-named codec emitted by the cbor_helpers template.
 
 import std/[os, strutils, tables, sets]
-import ./meta, ./string_helpers, ./c_cpp_common
+import ./meta, ./string_helpers, ./c_cpp_common, ./types_ir
 
 ## Wire-format C type for any Nim `ptr T` / `pointer`. Fixed 64-bit so the CBOR
 ## payload size is stable regardless of host architecture (mirrors CppPtrType).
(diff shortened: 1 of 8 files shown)


ffi/codegen/types_ir.nim
```diff
@@ -0,0 +1,134 @@
+## Structured type IR shared by the C / C++ / Rust binding generators: one
```
Ivansete-status: What is IR (infra red?) ?

ffi/codegen/types_ir.nim
```diff
@@ -0,0 +1,134 @@
...
+## parser (`parseFFIType`) for the Nim type strings each backend used to slice
+## by hand, plus `renderNative` to walk the result into a backend's type string.
+
+import std/[strutils, options]
+
+type
+  ScalarKind* = enum
```
Ivansete-status: I like .pure. enums as they force more explicit usage.

ffi/codegen/types_ir.nim
```diff
@@ -0,0 +1,134 @@
...
+    optOf*: proc(elem: string): string {.noSideEffect, nimcall.}
+    structName*: proc(name: string): string {.noSideEffect, nimcall.}
+      ## nil ⇒ the user type name passes through unchanged
+
+func genericInnerType(typeName, prefix: string): string =
+  ## Inner type of a single-parameter generic `Prefix[Inner]`, e.g.
+  ## `genericInnerType("seq[int]", "seq[")` → `"int"`; "" if not that shape.
+  if typeName.startsWith(prefix) and typeName.endsWith("]"):
+    return typeName[prefix.len .. ^2]
+  ""
```
Ivansete-status: nit, I prefer always being explicit.
```suggestion
  return ""
```

ffi/codegen/types_ir.nim
```diff
@@ -0,0 +1,134 @@
...
+
+func scalarKind(t: string): Option[ScalarKind] =
+  ## Single source of truth for the scalar leaf set every backend shares.
+  case t
```
Ivansete-status, line 62: Maybe needs toLower and trim?

ffi/codegen/types_ir.nim
```diff
@@ -0,0 +1,134 @@
...
+    return FFIType(kind: ftSeq, elem: parseFFIType(inner))
+
+  var optInner = genericInnerType(t, "Option[")
+  if optInner.len == 0:
+    optInner = genericInnerType(t, "Maybe[")
+  if optInner.len > 0:
+    return FFIType(kind: ftOpt, elem: parseFFIType(optInner.strip()))
+
+  let sc = scalarKind(t)
+  if sc.isSome:
```
Ivansete-status: nit, use parenthesis for verbs.
```suggestion
  if sc.isSome():
```

ffi/codegen/c.nim
```diff
@@ -293,44 +255,44 @@ proc emitStructType(reg: var CTypeReg, t: FFITypeMeta) =
...
-    return (leaf.cType, leaf.owns)
-
-  let seqInner = genericInnerType(t, "seq[")
-  if seqInner.len > 0:
-    let inner = seqInner.strip()
-    if inner == "byte" or inner == "uint8":
-      return ("NimFfiBytes", true)
-    let (elemC, _) = ensureCType(reg, inner)
+proc ensureCType(reg: var CTypeReg, t: FFIType): tuple[cType: string, owns: bool] =
+  ## Walks the shared type IR into a C type, monomorphising each distinct
```
Ivansete-status, line 259: We need to explain what does _own_ mean.

ffi/codegen/c.nim
```diff
@@ -293,44 +255,44 @@ proc emitStructType(reg: var CTypeReg, t: FFITypeMeta) =
...
+  ## `seq[T]` / `Option[T]` into its own struct + codec triple on first sight.
+  case t.kind
+  of ftPtr:
+    (CPtrType, false)
```
Ivansete-status: Let me repeat myself but I prefer being explicit on each returned type, if it is possible.
```suggestion
    return (CPtrType, false)
```

Ivansete-status (approved):
LGTM! Great simplification indeed! 🙌

ffi/codegen/types_ir.nim
gmelodie: Intermediate Representation. I had to look it up as well, claude stuff. will simplify

Issue #91, opened 2026-06-24
Add a native C-wire ABI + CBOR-free scalar fast path


gmelodie:
## Problem

Every call pays a full CBOR round-trip: encode req → `FFIThreadRequest.data` → handler → `handleRes` → callback with CBOR bytes. Scalar procs (e.g. `myTimerVersion -> string`, `int`/`bool` returns) pay CBOR for a single value.

`ABIFormat` (`ffi/codegen/meta.nim`) already defines `Cbor` and `C`, and the `C` path has a *type codec* (`*_CWire` companions + `ffi/internal/c_wire.nim`) but **no proc-dispatch path**. `abiCodegenImplemented(fmt)` is true only for `Cbor`; `gateABIFormat`/`gateFFITypeABIFormat` hard-error on `C` — the explicitly-marked seam to flip.

## Proposal (ship the two halves independently)

**2b — native scalar fast path (lower risk, ship first).** For procs whose entire signature + return is scalar (`int`/`intXX`/`float`/`bool`/`cstring`/handle-`uint64`): emit a direct C export `T_<proc>(ctx, args...) -> ret` with no envelope, no `_CWire`, no CBOR. Payload is a fixed-size POD inlined in `FFIThreadRequest` (union with `data`), so no `c_malloc` per call. Add `proc isScalarOnly(p: FFIProcMeta): bool` to `meta.nim` (pure compile-time over `extraParams`/`returnTypeName`).

**2a — C-wire dispatch (flat-struct ABI, larger).** Per `abi = c` proc, generate `<Proc>Req_CWire`/response structs via the existing codec; carry a `ptr <Req>_CWire` across the channel; add an `abi` discriminant to `FFIThreadRequest` so `processRequest` knows to CBOR-decode vs cast. `handleRes` returns the packed response pointer; foreign side reads fields then calls the generated `free_<Resp>`. Flip `abiCodegenImplemented(C)` → true and drop the `gateABIFormat` hard-error once it builds end to end.

## Gotchas

- **Ownership across the struct boundary** — `_CWire` cstrings are `cwireAllocStr`'d; foreign side must call the generated `free_*`, mirroring the CBOR `free_*_result` discipline.
- **ABI stability** — native structs churn the header on field changes; keep CBOR the default, native opt-in per proc.
- Pin struct layout (`{.packed.}` or explicit padding); assert `sizeof` parity (see parity harness).
- Don't regress CBOR: every existing proc stays `Cbor` unless annotated.

[...]

PR #106, opened 2026-07-01
feat(ffi): CBOR-free scalar fast path for all-scalar `abi = c` procs

gmelodie:
## Summary

Implements part **2b** of #91: a native, CBOR-free dispatch path for `{.ffi.}` methods whose entire signature is scalar. Today every call — even `myTimerVersion() -> string` or an `int`-returning proc — pays a full CBOR round-trip (encode args → malloc'd payload → decode → encode result → decode). For all-scalar procs that overhead is pure waste. This adds the opt-in fast path that skips it, reusing the existing FFI-thread queue, dispatch table, and callback machinery unchanged.

## Why this shape

- **Opt-in via the existing ABI seam.** A proc marked `{.ffi: "abi = c".}` whose params and return are all scalar takes the fast path; everything else stays on CBOR. No new pragma — this is the `abi = c` seam #91 pointed at. A non-scalar `abi = c` proc stays hard-gated (full C-wire dispatch is 2a, a separate PR).
- **Zero per-call payload allocation.** Scalar args are bit-cast into inline `uint64` slots on `FFIThreadRequest` (`ffiPackScalar`), so there's no `c_malloc`'d request buffer and no CBOR. The result rides back as raw bytes (`ffiScalarRetBytes`) — a POD scalar as its 8-byte little-endian image, a `string` as its own UTF-8 bytes (like the error path already does).
- **Inline array, not a `union` with `data`.** #91 suggested unioning the POD payload with `data` to avoid growing the struct. I used a plain `array[MaxScalarArgs, uint64]` instead: it costs a fixed 64 bytes per request but keeps the intrusive `next` link and `deleteRequest` unaliased and branch-free. Deliberate deviation, documented at the field.
- **Scalar logic lives in its own module.** Per review, the whole fast path is isolated in `ffi/internal/ffi_scalar.nim` (predicates + codegen). The base `ffi` macro just decides eligibility and delegates, so the usual CBOR/async path stays unaltered and simple.

## What's covered

- `ffi/internal/ffi_scalar.nim` (new): scalar-type predicates, `isScalarOnly(FFIProcMeta)`, `bindableProcs`, and `buildScalarPath` — the CBOR-free codegen. `ffi/codegen/meta.nim` keeps only the `scalarFastPath` registry flag it structurally has to.
- Inline scalar payload + `initScalar` (bounds-asserted) + the pack/unpack/serialize helpers (`ffi/ffi_thread_request.nim`).
- The `.ffi.` macro's scalar branch is a thin delegation into `buildScalarPath`, handing over the same shared dispatch pieces the usual path builds (`buildAsyncHelperProc` / `buildCtxGuard` / `buildSendAndReply`); no inline pack/unpack machinery in `ffi_macro.nim`.
- `genBindings` drops scalar procs via `bindableProcs` so the foreign generators don't emit a broken CBOR caller for a C export that no longer takes `(reqCbor, reqCborLen)`.

## Testing

- New `tests/unit/test_scalar_fastpath.nim`: C-export round-trips (two-int→int, no-param→string, float, bool, error→`RET_ERR`, invalid-ctx rejection), the Nim-native callable shape, and compile-time metadata assertions.
- Passes under `--mm:orc` and `--mm:refc`; the existing suite stays green and the committed example bindings regenerate byte-identically (no scalar procs in the examples, so foreign codegen is unaffected).

## Deliberately deferred (keeps this PR to the low-risk half of #91)

- **Foreign codegen** (Rust/C++/CDDL) for scalar procs — currently skipped, so the fast path isn't callable from Rust/C++ yet. Natural follow-up.
- **e2e C++ timer variant and the `bench_codec` cbor-vs-native latency numbers** from #91's test list — not added; the path is proven at the Nim unit level only.
- **cstring/string *params* and handle params/returns** — excluded from eligibility by design (a cstring param would need a copy to survive the async hop, defeating the zero-alloc promise). String *returns* are supported.
- **Empty-string scalar return** currently rides the shared `handleRes` CBOR-null sentinel; documented as a caveat the future foreign scalar reader must special-case.

### DIFF
--- a/ffi/internal/ffi_macro.nim
+++ b/ffi/internal/ffi_macro.nim
@@ -4,6 +4,7 @@ import ../ffi_types
 import ../ffi_thread_request
 import ../codegen/[meta, string_helpers]
 import ./c_macro_helpers
+import ./ffi_scalar
 when defined(ffiGenBindings):
   import ../codegen/rust
   import ../codegen/cpp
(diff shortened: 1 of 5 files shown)


ffi/internal/ffi_macro.nim
```diff
@@ -923,74 +1004,121 @@ macro ffi*(args: varargs[untyped]): untyped =
...
-          kind: FFIKind.FFI,
-          libTypeName: $libTypeName,
-          extraParams: ffiExtraParams,
-          returnTypeName: retTn,
-          returnIsPtr: retIsPtr,
-          returnIsHandle: retIsHandle,
-          abiFormat: abiFormat,
+    return newStmtList(helperProc, registerReq, ffiProc)
+
+  proc scalarPath(): NimNode =
```
Ivansete-status, line 1016: Given that "scalar" is an edge case, I wonder if we could have a separate "scalar" module so that we leave the usual path unaltered.

Ivansete-status (approved):
I wonder if considering the scalar path actually pays off the additional complexity. I think is better having something simpler even though is not extremely optimized for edge cases.
That's great for now but maybe we can isolate such _scalar_ logic so that the base logic remains relatively simple.
Thanks!

PR #107, opened 2026-07-01
feat(ffi): slab-allocate event payloads instead of c_malloc per dispatch

### LINKED ISSUE #92: Slab-allocate event payloads instead of c_malloc per dispatch
## Problem

`dispatchFFIEvent`/`dispatchFFIEventCbor` (`ffi/ffi_events.nim`) `c_malloc` a fresh payload buffer per event and `alloc()` the name string; the event thread `c_free`s both (`freeEventBuffers`). So every event = 2 mallocs + 2 frees, churning the global allocator under high-rate emission. Names are almost always a small fixed set of literals.

## Proposal (three independent wins, smallest first)

**3a — intern event names (near-zero risk, do first).** Event names are compile-time-known (`FFIEventMeta.wireName`). Store a `cstring` into a static/interned table instead of `alloc(evtName)` per dispatch; make `QueuedEvent.name` non-owning and stop freeing it. Removes one malloc + one free per event.

**3b — payload slab.** Pair each `EventQueue` slot with a preallocated buffer: add `slab: array[EventQueueCapacity, ptr UncheckedArray[byte]]` + `slabCap` sized to a `maxPayloadBytes` budget at `initEventQueue`. `tryEnqueueEvent` copies into `slab[tail]` (growing only that slot if exceeded); dequeue no longer frees; `deinitEventQueue` frees the slab once.

**3c — oversize fallback.** If a payload exceeds `maxPayloadBytes`, fall back to per-event `c_malloc` and tag the slot `heapOwned: bool` (free on dequeue). Keeps the slab small without sizing for the worst case.

Make `maxPayloadBytes` and `EventQueueCapacity` configurable per library (constructor / `declareLibrary` option) so embedded targets shrink and high-rate targets grow.

## Gotchas

- Single-producer (FFI thread enqueues) / single-consumer (event thread dequeues) — slab reuse is safe because a slot isn't reused until the ring wraps; keep the `count >= EventQueueCapacity` full check.
- Slab buffers must be `c_malloc`-backed (process-global), not `allocShared` — same TLS-region hazard as `ffi_thread_request.nim`.
- Don't break the `eventQueueStuck` sticky-overflow + `NotResponding` liveness path — overflow still drops + marks stuck, now without a leak.

## Tests

- Keep `test_event_dispatch.nim`, `test_event_thread.nim`, `test_event_listener.nim` green.
- High-rate emit (10^5 events) asserting zero steady-state mallocs (counting `c_malloc` shim, or massif / malloc-count interpose).
- asan for the oversize fallback, tsan for the producer/consumer handoff.

Effort: small for 3a, medium for 3b/3c.

gmelodie:
## Summary

Closes #92.

Under high-rate event emission every `dispatchFFIEvent` / `dispatchFFIEventCbor` cost two `c_malloc`s and two `c_free`s — one for the payload, one for the name string — churning the process-global allocator on the hot path. This replaces the per-event allocations with preallocated, reused per-slot buffers so steady-state emission allocates nothing.

Each `EventQueue` ring slot now owns two `c_malloc`-backed buffers, allocated once in `initEventQueue` and freed once in `deinitEventQueue`:

- a payload buffer sized to `MaxEventPayloadBytes` (default 512), and
- a name buffer sized to `MaxEventNameBytes` (default 64).

`tryEnqueueEvent` copies the name and payload into the tail slot's buffers instead of adopting caller-owned `c_malloc` pointers. A value that overflows its slot budget falls back to a one-off `c_malloc` tagged `nameHeapOwned` / `dataHeapOwned` and freed on commit, so the slab stays small without sizing for the worst case. Buffers remain `c_malloc`-backed (not `allocShared`) so the event thread can read a slot after the producing FFI thread's heap is gone — the same TLS hazard documented in `alloc.nim`.

`EventQueueCapacity`, `MaxEventPayloadBytes`, and `MaxEventNameBytes` are `{.intdefine.}` consts, overridable per built library (`-d:MaxEventPayloadBytes=N`, etc.).

## Two intentional divergences from the issue

1. The issue's 3a proposed interning names and never freeing them. That leaks under this repo's LSan CI (`detect_leaks=1`), so instead each slot reuses a name buffer freed at `deinit` — still allocation-free in steady state, but leak-clean. A leak-free process-global intern would require cross-thread-mutable GC state, which this codebase deliberately avoids.
2. The issue states slab reuse is safe "with the full check." It isn't under a stalled consumer plus a full ring: the single producer can overwrite a slab slot while a slow listener is still reading it. The drain path is now peek → dispatch → commit (`peekEvent` / `commitDequeue`), keeping the in-flight slot counted (and thus un-reusable) until dispatch returns. This is what keeps the producer/consumer handoff TSan-clean.

## Tests

- `test_event_dispatch.nim`, `test_event_thread.nim`, `test_event_listener.nim` stay green (orc and refc).
- Added `oversize payload falls back to heap` and `oversize event name falls back to heap`, covering both heap-fallback branches end-to-end.
- Verified locally: orc + refc pass; the oversize path is ASan/UBSan/LSan clean (`detect_leaks=1`); the overflow/backpressure handoff in `test_event_thread.nim` is TSan clean.

### DIFF
--- a/ffi/event_thread.nim
+++ b/ffi/event_thread.nim
@@ -61,17 +61,25 @@ proc onResponding*(ctx: ptr FFIContext) =
   emitLivenessEvent(ctx, RespondingEventName, RespondingEvent())
 
 proc dispatchQueuedEvent[T](ctx: ptr FFIContext[T], qe: QueuedEvent) =
-  ## Frees `qe`'s c_malloc buffers on exit.
-  defer:
-    freeEventBuffers(qe.name, qe.data)
+  ## Reads the borrowed slab payload; `commitDequeue` (not this proc) frees any
+  ## heap-fallback buffer once the read has returned.
   ctx.dispatchToListeners($qe.name, qe.data, qe.dataLen)
 
+proc drainOneEvent[T](ctx: ptr FFIContext[T]): bool =
+  ## Peek → dispatch → commit for a single event. The slot stays pinned across
+  ## dispatch so the producer can't reuse its slab buffer mid-read; `defer`
+  ## commits even if a listener raises. Returns false when the queue is empty.
+  let opt = ctx.eventQueue.peekEvent()
+  if opt.isNone():
+    return false
+  defer:
+    ctx.eventQueue.commitDequeue()
+  ctx.dispatchQueuedEvent(opt.get())
+  true
+
 proc drainEventQueue[T](ctx: ptr FFIContext[T]) =
-  while true:
-    let opt = ctx.eventQueue.tryDequeueEvent()
-    if opt.isNone():
-      break
-    ctx.dispatchQueuedEvent(opt.get())
+  while ctx.drainOneEvent():
+    discard
 
 type HeartbeatMonitor = object
   startedAt: Moment
@@ -125,7 +133,8 @@ proc eventRun[T](ctx: ptr FFIContext[T]) {.async.} =
 
 proc eventThreadBody[T](ctx: ptr FFIContext[T]) {.thread.} =
   ## Drains the event queue and runs the FFI-thread heartbeat check.
-  ## Owns the queued `c_malloc` payloads until dispatch returns.
+  ## Borrows each queued slab payload until dispatch returns, then releases
+  ## any heap-fallback buffer.
   defer:
     let fireRes = ctx.eventThreadExitSignal.fireSync()
     if fireRes.isErr():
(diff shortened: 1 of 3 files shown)


Ivansete-status (approved):
LGTM! Thanks for it! 🙌 
With regards to the event management, we need to invest efforts in the following, which basically aims to avoid dropping events (avoid fixed-sized ring) while keeping memory steady: https://github.com/logos-messaging/nim-ffi/issues/112

PR #108, opened 2026-07-01
feat(ffi): configurable per-request handler timeout with a finite default (#93)

### LINKED ISSUE #93: Add a configurable request timeout with a finite default
## Problem

`sendRequestToFFIThread(ctx, ffiRequest, timeout = InfiniteDuration)` (`ffi/ffi_thread.nim`) defaults to an infinite wait, so a wedged/slow handler blocks a foreign caller forever. The liveness watchdog (`NotRespondingEvent`/`RespondingEvent` via `ctx.ffiHeartbeat`) *detects* a wedged thread but the in-flight request still hangs — detection ≠ unblocking. Note the synchronous wait is only on *accept*, not *completion* (the result returns async via the callback), so there are two distinct timeouts.

## Proposal

**4a — submit/accept timeout.** Replace `InfiniteDuration` with a configurable `defaultRequestTimeout` on `FFIContext[T]` (default 5 s). On accept-timeout, return `err("ffi accept timeout")`. This turns the "FFI thread owns the request, don't double-free" note into a live race (the thread may pick the request up after the caller gave up): tag the request `abandoned: Atomic[bool]`; whoever sets it last frees it. Best merged with the MPSC per-request-completion work (#01).

**4b — handler/completion timeout (the one that matters).** Wrap the awaited handler in `processRequest` with `withTimeout(deadline)`; on expiry fire `RET_ERR` with a timeout message so the foreign side unblocks deterministically instead of waiting on the watchdog. Allow a per-proc override parsed like the existing `abi = ...` spec in `meta.nim` (e.g. `{.ffi: "timeout = 30000".}`; runtime-only, codegen ignores it).

## Gotchas

- **Cancellation safety** — `cancelSoon` on a handler mid-FFI to the underlying lib (Waku/libp2p) may leave it partial. Default to *report timeout but let the handler finish*; make hard-cancel an opt-in flag.
- The abandoned-request double-free race (4a) is the sharp edge — gate landing on a tsan run with artificial handler delays.
- 5 s default is a guess; make it trivially overridable, log at WARN on trip.

## Tests

- Handler that `sleepAsync`s past the deadline → caller gets timeout `err`, callback fires once, request freed.
- Handler finishing just under the deadline → normal path.
- tsan: concurrent submit + timeout + late FFI-thread pickup (abandoned claim).
- Keep `test_ffi_context.nim` liveness/heartbeat tests green.

Effort: medium. 4b is the high-value, lower-risk half; 4a's ownership race is best merged with #01.

gmelodie:
## Summary

Closes #93.

`sendRequestToFFIThread` used to hand a foreign caller an unbounded wait, so a wedged or slow handler blocked that caller forever — the liveness watchdog *detected* the wedge but the in-flight request still hung. This wires a finite, configurable handler-completion timeout so the caller unblocks deterministically.

## What changed

Each `FFIContext` now carries `defaultRequestTimeout` (5s, `ffi/ffi_context.nim`). In `processRequest` (`ffi/ffi_thread.nim`) the handler is raced against its deadline via `reportTimeoutIfTripped`; on trip the caller is answered with `ffi request timed out after <n>ms` and a WARN is logged.

The handler is deliberately **not** cancelled on timeout: a hard-cancel mid-call into the underlying library (Waku/libp2p) can leave it partially applied. We report the timeout and let the handler run to completion. `fireCallback` (`ffi/ffi_thread_request.nim`) carries a once-only `responded` guard so the two response paths (timeout vs. completion) fire the foreign callback **exactly once**, and freeing stays with `handleRes` so the request buffer has a single owner.

The deadline is overridable per proc with a `{.ffi: "timeout = <ms>".}` spec, parsed like the existing `abi = ...` spec (`meta.nim` `parseTimeoutSpec`/`overrideKey`, wired through `ffi`/`ffiRaw`/`ffiCtor` in `ffi_macro.nim`). It is runtime-only — recorded in `requestTimeoutsMs` at module init and read via the context; codegen ignores it.

## Notes / scope

The issue's 4a (submit/accept timeout) is obsolete on the current MPSC ingress: submission is a non-blocking enqueue with no accept-ack, so there is nothing to time out on submit. This PR implements 4b, the high-value half.

The timeout race sits inside the same `try` as the handler await, preserving the existing "`handleRes` must still run" cancellation-safety invariant. Already-completed handlers (e.g. sync bodies) skip arming a timer to keep the per-request cost off the hot path; a shared per-context timer to eliminate it entirely for async handlers is a possible follow-up. Hard-cancel-on-timeout is left as an opt-in follow-up per the issue's cancellation-safety note — a permanently-wedged handler still unblocks its caller but does not reclaim its own request buffer until it finishes.

## Tests

- `tests/unit/test_ffi_context.nim`: a handler past its per-proc deadline yields a timeout err and the callback fires exactly once (completion does not re-fire); a handler finishing under its deadline returns normally.
- `tests/unit/test_abi_format.nim`: `parseTimeoutSpec`/`overrideKey` parsing (valid, malformed, non-positive), and that a `timeout` override is recorded in `requestTimeoutsMs` while a plain proc has no entry.

Verified with the host nim-2.2.6 toolchain (`nim check` + compile-run of the unit suites); all unit tests check clean on orc and refc, and `nph` is idempotent on the touched files.

### DIFF
--- a/ffi/codegen/meta.nim
+++ b/ffi/codegen/meta.nim
@@ -67,19 +67,22 @@ proc abiCodegenImplemented*(fmt: ABIFormat): bool =
   ## seam a future PR flips once the `c` dispatch path is wired.
   fmt == ABIFormat.Cbor
 
-proc specKey*(spec: string): string =
-  ## Lowercased key of a `key = value` annotation spec (the text before `=`),
-  ## used to route a spec to its parser. `"timeout = 30000"` → `"timeout"`.
-  spec.split('=')[0].strip().toLowerAscii()
+proc overrideKey*(override: string): string =
+  ## Lowercased key of a `key = value` pragma override (the text before `=`),
+  ## used to route it to its parser. `"timeout = 30000"` → `"timeout"`.
+  override.split('=')[0].strip().toLowerAscii()
 
-proc parseTimeoutSpec*(spec: string): tuple[ok: bool, ms: int, err: string] =
+proc parseTimeoutSpec*(override: string): tuple[ok: bool, ms: int, err: string] =
   ## Parse a `"timeout = <milliseconds>"` override (whitespace/case tolerant).
   ## The value must be a positive integer number of milliseconds. On bad
   ## grammar or value, returns `ok = false` with a human-readable `err`.
-  let parts = spec.split('=')
-  if parts.len != 2 or specKey(spec) != "timeout":
-    return
-      (false, 0, "invalid timeout override '" & spec & "'; expected `timeout = <ms>`")
+  let parts = override.split('=')
+  if parts.len != 2 or overrideKey(override) != "timeout":
+    return (
+      false,
+      0,
+      "invalid timeout override: '" & override & "'; expected `timeout = <ms>`",
+    )
   let raw = parts[1].strip()
   let ms =
     try:
@@ -88,7 +91,7 @@ proc parseTimeoutSpec*(spec: string): tuple[ok: bool, ms: int, err: string] =
       return (
         false,
         0,
-        "invalid timeout value '" & raw &
+        "invalid timeout value: '" & raw &
           "'; expected a positive integer of milliseconds",
       )
   if ms <= 0:
@@ -106,28 +109,28 @@ proc parseABIFormatName*(name: string): tuple[ok: bool, fmt: ABIFormat] =
   else:
     (false, ABIFormat.Cbor)
 
-proc parseAbiSpec*(spec: string): tuple[ok: bool, fmt: ABIFormat, err: string] =
+proc parseAbiSpec*(override: string): tuple[ok: bool, fmt: ABIFormat, err: string] =
   ## Parse an `"abi = <format>"` override (whitespace/case tolerant). On bad
   ## grammar or format, returns `ok = false` with a human-readable `err`.
-  let parts = spec.split('=')
+  let parts = override.split('=')
   if parts.len != 2:
     return (
       false,
       ABIFormat.Cbor,
-      "invalid ABI override '" & spec & "'; expected `abi = c` or `abi = cbor`",
+      "invalid ABI override: '" & override & "'; expected `abi = c` or `abi = cbor`",
     )
   if parts[0].strip().toLowerAscii() != "abi":
     return (
       false,
       ABIFormat.Cbor,
-      "invalid ABI override '" & spec & "'; expected `abi = c` or `abi = cbor`",
+      "invalid ABI override: '" & override & "'; expected `abi = c` or `abi = cbor`",
     )
   let (ok, fmt) = parseABIFormatName(parts[1])
   if not ok:
     return (
       false,
       ABIFormat.Cbor,
-      "unknown ABI format '" & parts[1].strip() & "'; valid values are `c` and `cbor`",
+      "unknown ABI format: '" & parts[1].strip() & "'; valid values are `c` and `cbor`",
     )
   (true, fmt, "")
 
--- a/ffi/codegen/meta.nim
+++ b/ffi/codegen/meta.nim
@@ -67,6 +67,34 @@ proc abiCodegenImplemented*(fmt: ABIFormat): bool =
   ## seam a future PR flips once the `c` dispatch path is wired.
   fmt == ABIFormat.Cbor
 
+proc specKey*(spec: string): string =
+  ## Lowercased key of a `key = value` annotation spec (the text before `=`),
+  ## used to route a spec to its parser. `"timeout = 30000"` → `"timeout"`.
+  spec.split('=')[0].strip().toLowerAscii()
+
+proc parseTimeoutSpec*(spec: string): tuple[ok: bool, ms: int, err: string] =
+  ## Parse a `"timeout = <milliseconds>"` override (whitespace/case tolerant).
+  ## The value must be a positive integer number of milliseconds. On bad
+  ## grammar or value, returns `ok = false` with a human-readable `err`.
+  let parts = spec.split('=')
+  if parts.len != 2 or specKey(spec) != "timeout":
+    return
+      (false, 0, "invalid timeout override '" & spec & "'; expected `timeout = <ms>`")
+  let raw = parts[1].strip()
+  let ms =
+    try:
+      parseInt(raw)
+    except ValueError:
+      return (
+        false,
+        0,
+        "invalid timeout value '" & raw &
+          "'; expected a positive integer of milliseconds",
+      )
+  if ms <= 0:
+    return (false, 0, "timeout must be a positive number of milliseconds, got: " & raw)
+  (true, ms, "")
+
 proc parseABIFormatName*(name: string): tuple[ok: bool, fmt: ABIFormat] =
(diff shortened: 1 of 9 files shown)


ffi/codegen/meta.nim
```diff
@@ -67,6 +67,34 @@ proc abiCodegenImplemented*(fmt: ABIFormat): bool =
   ## seam a future PR flips once the `c` dispatch path is wired.
   fmt == ABIFormat.Cbor
 
+proc specKey*(spec: string): string =
+  ## Lowercased key of a `key = value` annotation spec (the text before `=`),
```
Ivansete-status: What is `spec`? Maybe better to use `configItem` term or `pragmaModifier`?

ffi/codegen/meta.nim
```diff
@@ -67,6 +67,34 @@ proc abiCodegenImplemented*(fmt: ABIFormat): bool =
...
+  spec.split('=')[0].strip().toLowerAscii()
+
+proc parseTimeoutSpec*(spec: string): tuple[ok: bool, ms: int, err: string] =
+  ## Parse a `"timeout = <milliseconds>"` override (whitespace/case tolerant).
+  ## The value must be a positive integer number of milliseconds. On bad
+  ## grammar or value, returns `ok = false` with a human-readable `err`.
+  let parts = spec.split('=')
+  if parts.len != 2 or specKey(spec) != "timeout":
+    return
+      (false, 0, "invalid timeout override '" & spec & "'; expected `timeout = <ms>`")
```
Ivansete-status: nit, I think is interesting to end the hard-coded strings with colon so that is easier to lookup.
```suggestion
      (false, 0, "invalid timeout override: '" & spec & "'; expected `timeout = <ms>`")
```

ffi/internal/ffi_macro.nim
```diff
@@ -604,7 +648,8 @@ macro ffiRaw*(args: varargs[untyped]): untyped =
 
   requireLibraryDeclared("`.ffiRaw.`")
   let prc = args[^1]
-  gateABIFormat(resolveABIFormat(args[0 ..^ 2]), "`.ffiRaw.` proc")
+  let (rawAbiFormat, rawTimeoutMs) = resolveFFISpecs(args[0 ..^ 2])
```
Ivansete-status, line 653: Maybe needed to trim?

ffi/internal/ffi_macro.nim
```diff
@@ -747,14 +793,16 @@ macro ffi*(args: varargs[untyped]): untyped =
...
   let prc = args[^1]
-  let abiFormat = resolveABIFormat(args[0 ..^ 2])
+  let (abiFormat, timeoutMs) = resolveFFISpecs(args[0 ..^ 2])
 
   # A value type stands alone (no library required). Its `c` companion is
   # emitted later by `genBindings()`, since a type-pragma macro can only return
   # a TypeDef; `cbor` rides the generic overloads. Both abis are valid here.
   if prc.kind == nnkTypeDef:
+    if timeoutMs > 0:
+      error("`.ffi.` on a type takes no `timeout` override (it applies to procs)")
```
Ivansete-status:
```suggestion
      error("`.ffi.` on a type takes no `timeout` override (it only applies to procs)")
```

ffi/ffi_context.nim
```diff
@@ -39,6 +39,15 @@ type FFIContext*[T] = object
...
   registeredRequests: ptr Table[cstring, FFIRequestProc]
+  requestTimeouts: ptr Table[cstring, int]
+    # Per-proc timeout overrides (ms). Points at the compile-time-filled global,
+    # like registeredRequests, so the FFI thread reads it GC-safely via ctx.
+  defaultRequestTimeout*: Duration
+    # Deadline applied to each handler unless a `{.ffi: "timeout = <ms>".}`
+    # override raises it. On trip the caller is unblocked with a timeout err and
+    # the handler is left running (see processRequest). Set `InfiniteDuration`
+    # to opt out. Written on the owning thread before the first request; read on
+    # the FFI thread.
```
Ivansete-status: Let's have these comments more concise or ideally, one liner.

ffi/ffi_context.nim
```diff
@@ -49,6 +58,9 @@ const
   EventThreadTickInterval* = 1.seconds
   FFIHeartbeatStartDelay* = 10.seconds # grace window for library startup
   FFIHeartbeatStaleThreshold* = 1.seconds
+  DefaultRequestTimeout* = 5.seconds
+    # A guess (issue #93): finite so a wedged handler can't hang a caller
+    # forever, generous enough to clear normal handlers. Overridable per proc.
```
Ivansete-status:
```suggestion
```

ffi/ffi_thread_request.nim
```diff
@@ -125,12 +129,9 @@ proc deleteRequest*(request: ptr FFIThreadRequest) =
...
   c_free(request)
 
-proc handleRes*(res: Result[seq[byte], string], request: ptr FFIThreadRequest) =
-  ## Fires the registered callback exactly once and frees the request.
-  ## Success payload is CBOR bytes; error payload is the raw UTF-8 error string.
-  defer:
-    deleteRequest(request)
-
+proc fireCallback(res: Result[seq[byte], string], request: ptr FFIThreadRequest) =
+  ## Delivers one response to the foreign callback. Success payload is CBOR
```
Ivansete-status: I think better not to give explicit impl detail. I suggest to not to mention CBOR as it can be raw C type in some cases.

ffi/ffi_thread_request.nim
```diff
@@ -155,5 +156,23 @@ proc handleRes*(res: Result[seq[byte], string], request: ptr FFIThreadRequest) =
...
+    return
+  request[].responded = true
+  fireCallback(res, request)
+
+proc handleRes*(res: Result[seq[byte], string], request: ptr FFIThreadRequest) =
+  ## Terminal step of every request: delivers the response (unless a timeout
+  ## already did) and frees the request exactly once.
+  defer:
+    deleteRequest(request)
+  respondOnce(request, res)
```
Ivansete-status: Maybe?
```suggestion
  respondOnce(request, res)
  deleteRequest(request)
```

ffi/ffi_thread_request.nim
```diff
@@ -155,5 +156,23 @@ proc handleRes*(res: Result[seq[byte], string], request: ptr FFIThreadRequest) =
...
 
+proc respondOnce*(request: ptr FFIThreadRequest, res: Result[seq[byte], string]) =
+  ## Fires the callback the first time it's called for `request` and no-ops
+  ## after — the timeout path and the handler-completion path both call it, but
+  ## the foreign side must be answered exactly once. Does NOT free the request:
+  ## freeing stays with `handleRes` so the handler always owns the buffer until
+  ## it finishes.
+  if request[].responded:
+    return
+  request[].responded = true
```
Ivansete-status: Maybe this logic can go inside `fireCallback` and then we don't need the `respondOnce`.

Ivansete-status (approved):
LGTM! Thanks for it! 🙌 
Just added some nitpicks that I hope you find useful.

PR #109, opened 2026-07-01
feat: CBOR-free abi=c C bindings (-d:targetLang=c_abi)

### LINKED ISSUE #105: Emit ABI examples
Emit ABI examples such as `abi=c`, not only CDDL. https://github.com/logos-messaging/nim-ffi/pull/102#discussion_r3501788118

gmelodie:
## Summary

Closes #105. Follow-up to the C binding generator (#102), which emits a **CBOR** C binding (vendoring TinyCBOR). #105 asks for the *pure C ABI* rendering too — `abi=c`, no CBOR — so the two can be compared side by side.

This adds `-d:targetLang=c_abi`: a **CBOR-free** C backend whose flat structs *are* the ABI, and wires the `abi=c` proc-dispatch path end-to-end (previously only the `_CWire` type codecs existed; procs/ctors were still gated).

The C consumer surface is CBOR-free; CBOR remains only as an *internal* transport, invisible to the consumer:
- the generated `<lib>.h` is self-contained (`#include <stdint.h>` etc.) — **no TinyCBOR, no `nim_ffi_cbor.h`** — and its structs mirror the macro's `_CWire` layout byte-for-byte (`ShoutRequest { const char* text; }`);
- generated Nim exported wrappers `cwireUnpack` the incoming struct → Nim object, reuse the existing CBOR thread transport, and a Nim reply trampoline `cwirePack`s the response back into a flat struct for the caller's typed callback.

`examples/echo/c_abi_bindings/` is checked in beside the CBOR `examples/echo/c_bindings/` as the comparison artifact (see its README for the table).

## How it works

- `abiCodegenImplemented` now accepts `c` for proc/ctor/dtor annotations (`c` events stay CBOR-only, rejected with a clear compile error).
- `ffi/internal/c_macro_helpers.nim` grows the `abi=c` dispatch generator + a compile-time spec registry. All dispatch code (exported wrappers, reply trampolines, per-proc Req `_CWire` companions) is emitted at `genBindings()` time, after `flushCWireCompanions`, so nested companions and their `cwireUnpack`/`cwirePack` overloads are already in scope.
- `ffi/codegen/c_abi.nim` (new) is the CBOR-free header/CMake generator. Its struct layout stays in lock-step with `wireValueType`/`wireFieldsFor`.
- `examples/echo/echo.nim` gains a `-d:ffiEchoAbiC` toggle (flips `declareLibrary` to `defaultABIFormat = "c"`) so one source drives both the CBOR and flat example outputs. A library that always wants the C ABI just sets that in its own `declareLibrary`.

## Callback contract (null-safety)

- The reply callback's `err_msg` is **always a non-nil C string** — empty (`""`) on the success paths — so a consumer can print/`strlen` it unconditionally.
- For a **string return**, the `reply` argument is likewise always non-nil (empty string on the error paths).
- For an **object return**, `reply` is a `const <Struct>*` that is nil only on error, gated by a non-`RET_OK` `err_code` (the C-idiomatic "no value"); consumers check `err_code` before dereferencing, exactly as the e2e test does.
- The generated create trampoline null-checks its box before dereferencing, and `NIMFFI_RET_*` are plain `#define`s (identical across headers, so multi-lib inclusion is a benign redefinition).
- Nothing crossing the boundary is consumer-owned: request strings are borrowed for the call, reply structs/strings are freed by the binding after the callback returns, and the only teardown the consumer performs is `<lib>_ctx_destroy`.

## Testing

- `tests/unit/test_c_abi_codegen.nim` (8 cases) drives `generateCAbiLibHeader` against a synthetic registry — covers flat structs, the `seq[T]` `_items`/`_len` split, `Option[T]` pointers, string vs object returns, the ctor raw-address callback, and the events-rejected path.
- `tests/e2e/c_abi` ctest harness (`test_echo_c_abi.c`) exercises the real dispatch: ctor, object-returning method, string-returning method, teardown. C11 atomics on the done-flag keep it TSan-clean.
- New nimble tasks: `genbindings_c_abi_echo`, `check_bindings_c_abi` (now part of `check_bindings`), `test_c_abi_e2e`, `test_c_abi_e2e_sanitized`.
- CI: Linux `abi=c` e2e step in `ci.yml`; sanitized step in `tests-sanitized.yml`.
- Validated locally: e2e passes plain and clean under **ASan+UBSan and TSan**; 240 stress runs (no leaks/UAF/races). No regressions — all pre-existing unit tests, the CBOR timer C e2e, and full `check_bindings` (rust/cpp/c/c_abi) still pass.

## Scope notes

- Object and string returns are wired (echo's full surface). Handle/void/tuple/array returns and `abi=c` events raise a clear codegen error for a later slice.
- Only the echo example is emitted as `abi=c` (timer uses a `{.ffiEvent.}`, which `abi=c` doesn't support yet).
- The runtime `abi=c` e2e runs Linux-only (the generated header is platform-checked separately); the flat-struct layout is compiler/ABI-independent, so one OS exercising the round-trip is sufficient.


examples/echo/c_abi_bindings/echo.h
```diff
@@ -0,0 +1,122 @@
...
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+
+#ifndef NIMFFI_RET_OK
+#define NIMFFI_RET_OK 0
+#define NIMFFI_RET_ERR 1
+#define NIMFFI_RET_MISSING_CALLBACK 2
+#endif
+
```
Ivansete-status: This is a bit weird and seems not needed.
```suggestion
#define NIMFFI_RET_OK 0
#define NIMFFI_RET_ERR 1
#define NIMFFI_RET_MISSING_CALLBACK 2

```

examples/echo/c_abi_bindings/echo.h
```diff
@@ -0,0 +1,122 @@
...
+/* High-level context wrapper */
+typedef struct {
+    void* ptr;
+} EchoCtx;
+
+typedef void (*EchoCreateFn)(int err_code, EchoCtx* ctx, const char* err_msg, void* user_data);
+typedef struct { EchoCreateFn fn; void* user_data; } EchoCreateBox;
+static void echo_create_trampoline(int ret, const char* ctx_addr, const char* err_msg, void* ud) {
+    EchoCreateBox* box = (EchoCreateBox*)ud;
+    if (!box->fn) { free(box); return; }
```
Ivansete-status, line 57: Maybe interesting box != NULL before free.

ffi/internal/c_macro_helpers.nim
```diff
@@ -553,3 +553,370 @@ proc flushCWireCompanions*(): NimNode {.compileTime.} =
...
+    defer:
+      freeBox(box)
+    if box.fn.isNil():
+      return
+    try:
+      if ret != RET_OK:
+        var em = newString(int(len))
+        if int(len) > 0:
+          copyMem(addr em[0], msg, int(len))
+        box.fn(ret, nil, em.cstring, box.ud)
```
Ivansete-status, line 698: I think is better to never return `nil` and instead, return a fixed empty string. That would avoid unexpected crashes in consumer's code. Applies elsewhere.

ffi/internal/c_macro_helpers.nim
```diff
@@ -553,3 +553,370 @@ proc flushCWireCompanions*(): NimNode {.compileTime.} =
...
+    if onReply.isNil():
+      return RET_MISSING_CALLBACK
+    if not `poolIdent`.isValidCtx(cast[pointer](ctx)):
+      onReply(RET_ERR, nil, "ctx is not a valid FFI context".cstring, userData)
+      return RET_ERR
+    var reqObj: `envName` = cwireUnpack(req[])
+    let enc = cborEncodeShared(reqObj)
+    let reqBuf = enc.data
+    let reqBufLen = enc.len
+    let box = cast[ptr `boxName`](allocBox(sizeof(`boxName`)))
```
Ivansete-status, line 771: We need to make sure the lib's consumer doesn't need to manually free anything. I think 99% this is the case with that and is just a helper to retrieve the lib result.

Ivansete-status (approved):
Nice! Thanks for it! 🙌 
Maybe we need to allow generating both ABI simultaneously too. I mean, add another default `abi = c_and_cbor`. We will see if needed in the future.

PR #110, opened 2026-07-01
feat: CBOR-free scalar fast path for abi=c procs

gmelodie:
## Summary

Adds a CBOR-free dispatch path for `{.ffi: "abi = c".}` procs whose entire signature is scalar, so a call like `add(ctx, cb, ud, a, b) -> int` crosses the FFI boundary without a CBOR encode/decode and without a per-call *payload* `c_malloc` on either leg. The motivation is the hot path: small scalar calls previously paid a full CBOR envelope round-trip (encode → `c_malloc` payload → decode) for what is ultimately a handful of machine words. The request envelope itself (and its routing `reqId`) is still `c_malloc`'d, exactly as on the CBOR path — this removes the *extra* payload allocation and the serialization, not every allocation.

## How it works

- `isScalarOnly` (`ffi/codegen/meta.nim`) decides eligibility at compile time: every wire param and the return type must be a fixed-width POD scalar (`isScalarParamTypeName`) — handles and raw pointers are excluded. A `string`/`cstring` **return** is allowed (the handler owns the bytes and they ride back raw, like the error path), but not as a param, since a string param aliases caller memory the FFI thread reads later.
- Eligible procs pack their args inline into `FFIThreadRequest.scalarArgs` (a fixed `array[MaxScalarArgs, uint64]`) via `ffiPackScalar`, so the request carries the payload by value — no envelope payload `c_malloc`. The FFI-thread handler unpacks with `ffiUnpackScalar`, runs the user body, and returns the result as raw bytes via `ffiScalarRetBytes` (a small `seq[byte]`, no CBOR).
- The macro (`ffi/internal/ffi_macro.nim`) grew a `scalarPath()` alongside the existing `asyncPath()`; shared machinery (`buildCtxGuard`, `buildSendAndReply`, `buildCExportProc`, `wireParamMeta`, one `procMeta` built once) was extracted so both paths reuse it instead of duplicating the C-export/enqueue boilerplate.
- Scalar procs are dropped from `bindableProcs`, so all four foreign-binding generators (Rust/C++/C/CDDL) skip them instead of emitting a CBOR caller whose `(reqCbor, reqCborLen)` prototype wouldn't match the scalar proc's inline-arg export. `genBindings` emits a compile-time hint naming each skipped proc so the omission is visible. Foreign-binding codegen for the scalar shape is a deliberate follow-up; the metadata is still registered (with `scalarFastPath = true`) so it stays introspectable.

## Empty-string return fix

A scalar proc returning `""` serialises to a 0-length payload. The shared `handleRes` success branch treated *any* 0-length payload as "no value" and substituted the 1-byte CBOR-null sentinel (`0xf6`), so `""` reached the C wire as a 1-char string `"\xf6"`. `handleRes` is now scalar-aware: for a scalar request a 0-length result is delivered as a genuine 0-length `RET_OK` buffer (non-nil ptr, `len = 0`), which reads back as `""`. The CBOR path is untouched — `isScalar` is always false there, and a CBOR-encoded empty string is 1 byte, not 0, so the sentinel branch stays correct for genuine void returns.

## Tests

`tests/unit/test_scalar_fastpath.nim` covers both the C-export shape (driven through a real FFI thread) and the Nim-native shape (the proc stays directly callable as its declared `Future[Result[T, string]]`): int/float/bool round-trips, string return, the error path, invalid-ctx rejection, and the empty-string case whose regression guard asserts `d.msgLen == 0` (pre-fix it was `1`, the sentinel). A `static:` block asserts the scalar procs are flagged `scalarFastPath`, recognised by `isScalarOnly`, and dropped from `bindableProcs`. Green under both `--mm:orc` and `--mm:refc`.

### DIFF
--- a/ffi/ffi_thread_request.nim
+++ b/ffi/ffi_thread_request.nim
@@ -177,9 +177,9 @@ func ffiScalarRetBytes*[T](x: T): seq[byte] =
   ## Serializes a scalar handler result into the raw response payload the
   ## callback carries — no CBOR envelope. A `string`/`cstring` rides as its
   ## own UTF-8 bytes (like the error path); every other scalar rides as the
-  ## 8-byte native-endian image of `ffiPackScalar(x)`. Note: an empty string
-  ## yields a 0-length payload, which `handleRes` sends as the CBOR-null
-  ## sentinel — the foreign scalar reader (a follow-up) must special-case it.
+  ## 8-byte native-endian image of `ffiPackScalar(x)`. An empty string yields a
+  ## 0-length payload, which `handleRes` delivers as a genuine 0-length
+  ## `RET_OK` buffer (not the CBOR-null sentinel), so it reads back as "".
   when T is string:
     var b = newSeq[byte](x.len)
     if x.len > 0:
@@ -230,6 +230,13 @@ proc fireCallback*(res: Result[seq[byte], string], request: ptr FFIThreadRequest
         cast[ptr cchar](unsafeAddr bytes[0]),
         cast[csize_t](bytes.len),
         request[].userData,
+      )
+    elif request[].isScalar:
+      # A scalar return of 0 bytes is a real empty string, not "no value":
+      # hand back a genuine empty buffer, not the CBOR-null sentinel.
+      var empty: byte
+      request[].callback(
+        RET_OK, cast[ptr cchar](addr empty), 0.csize_t, request[].userData
       )
     else:
       # Always hand the callback a real buffer; CBOR null marks "no value".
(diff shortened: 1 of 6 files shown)


ffi/ffi_thread_request.nim
```diff
@@ -228,6 +228,13 @@ proc handleRes*(res: Result[seq[byte], string], request: ptr FFIThreadRequest) =
         cast[csize_t](bytes.len),
         request[].userData,
       )
+    elif request[].isScalar:
+      # A scalar return of 0 bytes is a real empty string, not "no value":
```
Ivansete-status: Is interesting to also remind what does request isScalar mean in this context.

Ivansete-status (approved):

PR #111, opened 2026-07-01
feat(devx): derive ffiEvent wire name, guard genBindings ordering, real README

gmelodie:
## Summary

Three developer-experience improvements to the FFI authoring surface, grounded in how a real consumer uses it (the `nim-libp2p` `cbind/` migration to `nim-ffi`, vacp2p/nim-libp2p#2681). No runtime or wire-format changes: existing libraries that already spell out explicit `{.ffiEvent.}` names and keep `genBindings()` last generate the same bindings as before.

### 1. `{.ffiEvent.}` wire name is now optional
Every event in real usage was written as `{.ffiEvent: "on_incoming_stream".}` where the string is exactly `camelToSnakeCase(procName)`. `{.ffi.}` already derives its C export symbol that way, so events were the odd one out. When the string is omitted it is now derived from the proc name (`onPeerConnected` → `on_peer_connected`); pass a literal only to override. Backward compatible — explicit names still work, and a lone `"abi = ..."` spec is now also accepted without a wire name.

### 2. Annotations after `genBindings()` are a loud error
`genBindings()` reads compile-time registries the pragmas populate as the compiler expands them, so an annotation placed after it never reached the generator and was **silently dropped** from the bindings — previously only warned about in a prose comment. A new `genBindingsEmitted` compile-time flag makes `{.ffi.}` / `{.ffiCtor.}` / `{.ffiDtor.}` / `{.ffiEvent.}` / `{.ffiHandle.}` / `{.ffiRaw.}` fail with a message pointing at the fix. The guard fires in every build (not only under `-d:ffiGenBindings`), so bad ordering is caught at normal compile time. Multi-file libraries are unaffected: imports expand before the root file body, so `genBindings()` at the root bottom still sees a fully-populated registry.

### 3. Real README
The README was a two-line stub. It now documents the mental model, a minimal end-to-end example, a pragma reference table, the `Future[Result[T, string]]` return-type contract, events (incl. the derived name), ABI formats, the `genBindings()` placement rule, and the codegen build flags.

## Testing

Verified against the real Nim 2.2.6 toolchain (the CI-pinned version):
- `nim check` on `ffi.nim` and the `examples/timer` library — clean.
- `tests/unit/test_event_listener`, `test_event_dispatch`, `test_meta` — pass.
- Generated C bindings from a no-arg `{.ffiEvent.}` — emit the derived `on_peer_connected` name in the trampoline, listener helper, and CBOR `eventType` string.
- An annotation placed after `genBindings()` — now fails compilation with the ordering error instead of silently vanishing.

## Notes

`{.ffiEvent.}` still supports exactly one payload parameter (unchanged). The enum-across-the-wire and C++ `seq[byte]`/`uint64` codegen gaps surfaced by the `cbind` consumer are separate follow-ups, not in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,10 +14,11 @@ No hand-written `.h` files, no manual request enums, no shared-memory plumbing.
   `{.ffiDtor.}`, `{.ffiEvent.}`).
 - You **call `genBindings()` last**, which emits the foreign bindings.
 
-Every request/response crosses the boundary as a single CBOR blob; the ctx
-handle returned by the constructor is the only pointer that crosses. Each
-`{.ffi.}` proc runs on the library's own chronos event loop, so bodies can
-`await` freely.
+By default, every request/response crosses the boundary as a single CBOR blob
+(the wire format is configurable per library or per annotation — see
+[ABI format](#abi-format)); the ctx handle returned by the constructor is the
+only pointer that crosses. Each `{.ffi.}` proc runs on the library's own chronos
+event loop, so bodies can `await` freely.
 
 ## Minimal example
 
@@ -31,7 +32,8 @@ type Counter = object
 declareLibrary("counter", Counter)
 
 # 2. Request/response shapes. Any {.ffi.} object type becomes a first-class
-#    struct/class in the generated bindings and rides the wire as CBOR.
+#    struct/class in the generated bindings and rides the wire in the library's
+#    ABI format (CBOR by default).
 type BumpRequest {.ffi.} = object
   by: int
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,7 +1,148 @@
...
+worker thread, a request channel, CBOR (de)serialization, an event queue and a
+context/handle registry — and generates the foreign-language bindings for you.
+No hand-written `.h` files, no manual request enums, no shared-memory plumbing.
+
+## Mental model
+
+- You **declare a library** once with `declareLibrary(name, LibType)`.
+- You **annotate procs and types** with pragmas (`{.ffi.}`, `{.ffiCtor.}`,
+  `{.ffiDtor.}`, `{.ffiEvent.}`).
+- You **call `genBindings()` last**, which emits the foreign bindings.
```
Ivansete-status: Maybe one possible future optimization is to only call declareLibrary _last_ and make it to also invoke `genBindings()` internally.

README.md
```diff
@@ -1,7 +1,148 @@
...
+
+Every request/response crosses the boundary as a single CBOR blob; the ctx
```
Ivansete-status: This is not necessarily true if the used ABI is pure C.

Ivansete-status (approved):
Great devx improvement. Thanks!

PR #115, opened 2026-07-08
feat(ffi): async teardown hook for {.ffiDtor.}

gmelodie:
## Summary

Lets a library run **async cleanup on the FFI worker thread** during context teardown, and makes `destroy` block until it finishes. This is what unblocks an `async` `libp2pDestroy` that `await`s `switch.stop()` so the switch is shut down gracefully — on the loop its chronos futures actually live on — before the FFI/event threads are joined and the context is freed.

Before this, a `{.ffiDtor.}` body ran **synchronously on the host thread**, with no seam to run async work on the worker at shutdown.

## Mechanism

A non-empty `{.ffiDtor.}` body is lifted into an async impl registered in a per-library **runtime hook slot** (`ffiTeardownHook[T]()`, one `{.global.}` per library type). After the FFI thread drains in-flight requests and before it exits, it `await`s the hook on its own event loop. Exceptions are logged, never propagated — the thread must still fire `threadExitSignal` and exit cleanly.

**Why a runtime slot and not an overloaded `ffiTeardown` resolved via `mixin`** (the obvious design): the constructor force-instantiates `ffiThreadBody[T]` at its `createFFIContext` call, which happens **before** the dtor — declared later in the source, as every consumer writes it (ctor then dtor) — is visible. An overload would bind the no-op default at that instantiation, so the teardown would compile clean, `destroy` would return `RET_OK`, and the body would **silently never run**. Confirmed empirically: natural ctor-before-dtor order skips the teardown; dtor-first runs it. The runtime slot is assigned at module init and read at destroy, with zero compile-time ordering coupling.

## Changes

- `ffi/ffi_context.nim` — `FFITeardownProc[T]` + `ffiTeardownHook[T]()` runtime slot. `ThreadExitTimeout` is now configurable via `-d:ffiThreadExitTimeoutMs=<ms>` (default unchanged at 1500): a graceful `switch.stop()` over many live connections can outlast the old fixed bound, and being cut short leaks the context instead of waiting.
- `ffi/ffi_thread.nim` — await the registered hook at shutdown, guarded on hook-set and `myLib`-populated.
- `ffi/internal/ffi_macro.nim` — `{.ffiDtor.}` now accepts a sync (`void`) or async (`Future[void]`) signature, validated up front. Non-empty bodies are lifted into the hook impl; the C wrapper is reduced to nil-check + `destroyFFIContext`. An empty/`discard` body registers no hook, so existing no-op dtors (echo, timer, waku) are unchanged.

## Testing

- `tests/unit/test_ffi_teardown.nim` (new) — asserts `destroy` blocks until the async teardown body completes, and that it runs on the FFI thread (not the caller). Green under both `--mm:orc` and `--mm:refc`.
- Verified back-compat: legacy sync-body dtors and the existing no-op `discard` dtors still compile and run; `nimble check_bindings_c` shows generated headers are unchanged; timer/echo examples still build; the `-d:ffiThreadExitTimeoutMs` knob compiles and takes effect.

## Rollout note

Consumer PR: vacp2p/nim-libp2p#2773 (https://github.com/vacp2p/nim-libp2p/pull/2773) re-pins nim-ffi to this commit and lands the async `libp2pDestroy` + shared `shutdownSwitch` + `stopped` guard.

That side should build with a larger `-d:ffiThreadExitTimeoutMs` — the default 1500 ms is likely too short for a graceful multi-connection `switch.stop()`, and a teardown that outlasts the bound leaks the context (returns `RET_ERR`) rather than hanging the caller.

### DIFF
--- a/ffi/ffi_context.nim
+++ b/ffi/ffi_context.nim
@@ -37,6 +37,9 @@ type FFIContext*[T] = object
   ffiHeartbeat*: Atomic[int64]
     # advanced each FFI-thread loop; event thread reads for liveness
   eventQueueStuck*: Atomic[bool] # sticky overflow flag
+  ffiThreadExited*: Atomic[bool]
+    # set once the FFI thread (including any async {.ffiDtor.} teardown) is done;
+    # keeps the event thread draining until then so teardown-emitted events land
   running: Atomic[bool] # To control when the threads are running
   registeredRequests: ptr Table[cstring, FFIRequestProc]
   requestTimeouts: ptr Table[cstring, int]
@@ -126,6 +129,7 @@ proc initContextResources*[T](ctx: ptr FFIContext[T]): Result[void, string] =
   initEventQueue(ctx[].eventQueue)
   ctx.ffiHeartbeat.store(0)
   ctx.eventQueueStuck.store(false)
+  ctx.ffiThreadExited.store(false)
   ctx.defaultRequestTimeout = DefaultRequestTimeout
 
   var success = false
@@ -199,7 +203,7 @@ proc signalStop*[T](ctx: ptr FFIContext[T]): Result[void, string] =
 ## connections) on the FFI thread before it exits — a graceful shutdown can
 ## outlast the default, and being cut short leaks the context instead of waiting.
 ## Override at compile time with `-d:ffiThreadExitTimeoutMs=<ms>`.
-const ThreadExitTimeoutMs* {.intdefine.} = 1500
+const ThreadExitTimeoutMs* {.intdefine: "ffiThreadExitTimeoutMs".} = 1500
 const ThreadExitTimeout* = ThreadExitTimeoutMs.milliseconds
 
 proc stopAndJoinThreads*[T](ctx: ptr FFIContext[T]): Result[void, string] =
--- a/ffi/ffi_context.nim
+++ b/ffi/ffi_context.nim
@@ -58,6 +58,19 @@ const
   DefaultRequestTimeout* = 5.seconds
     # Finite fallback (issue #93) so a wedged handler can't hang a caller forever.
 
+type FFITeardownProc*[T] = proc(lib: ptr T): Future[void] {.async.}
+
+proc ffiTeardownHook*[T](): var FFITeardownProc[T] =
+  ## Per-library teardown slot (one `{.global.}` per `T`), assigned at module init
+  ## by a non-empty `{.ffiDtor.}` and awaited by the FFI thread before it exits.
+  ##
+  ## A runtime slot, not an overloaded `ffiTeardown` resolved via `mixin`: the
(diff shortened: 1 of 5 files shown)


ffi/internal/ffi_macro.nim
```diff
@@ -1575,26 +1575,34 @@ macro ffiCtor*(args: varargs[untyped]): untyped =
...
+  ## live on the FFI event loop.
   ##
   ## The wire format follows the library default and can be overridden with
   ## `{.ffiDtor: "abi = c".}` / `{.ffiDtor: "abi = cbor".}`.
   ##
-  ## Example:
-  ##   proc waku_destroy*(w: Waku) {.ffiDtor.} =
-  ##     w.cleanup()
+  ## Example (async):
+  ##   proc waku_destroy*(w: Waku): Future[void] {.ffiDtor.} =
```
Ivansete-status, line 1593: May we also have a sync example to demonstrate the flexibility of both?

ffi/ffi_context.nim
```diff
@@ -58,6 +58,19 @@ const
   DefaultRequestTimeout* = 5.seconds
     # Finite fallback (issue #93) so a wedged handler can't hang a caller forever.
 
+type FFITeardownProc*[T] = proc(lib: ptr T): Future[void] {.async.}
+
+proc ffiTeardownHook*[T](): var FFITeardownProc[T] =
+  ## Per-library teardown slot (one `{.global.}` per `T`), assigned at module init
+  ## by a non-empty `{.ffiDtor.}` and awaited by the FFI thread before it exits.
```
Ivansete-status, line 68: I guess event thread should also wait for it before ending.

ffi/ffi_context.nim
```diff
@@ -180,9 +193,14 @@ proc signalStop*[T](ctx: ptr FFIContext[T]): Result[void, string] =
...
-## leaking ctx rather than hanging the caller.
-const ThreadExitTimeout* = 1500.milliseconds
+## Bound on how long stopAndJoinThreads waits for a worker thread to exit before
+## leaking ctx rather than hanging the caller. Configurable because an async
+## `{.ffiDtor.}` runs its teardown (e.g. `switch.stop()` over many live
+## connections) on the FFI thread before it exits — a graceful shutdown can
+## outlast the default, and being cut short leaks the context instead of waiting.
+## Override at compile time with `-d:ffiThreadExitTimeoutMs=<ms>`.
+const ThreadExitTimeoutMs* {.intdefine.} = 1500
+const ThreadExitTimeout* = ThreadExitTimeoutMs.milliseconds
```
Ivansete-status:
```suggestion
const ThreadExitTimeoutMs* {.intdefine: "ffiThreadExitTimeoutMs".} = 1500
```

Ivansete-status (approved):
LGTM! Thanks for it! 🙌

PR #116, opened 2026-07-09
feat(ffi): compile-time validation of --nimMainPrefix

gmelodie:
## Summary

`declareLibraryBase` imports the Nim runtime init symbol by name (using `importc`) as `lib{libraryName}NimMain`, so the build must be invoked with `--nimMainPrefix:lib{libraryName}` matching the `declareLibrary` name exactly. When the two disagree, the build fails at *link* time with an opaque undefined-symbol error that points nowhere near the actual mistake.

This moves that failure to compile time with a message that names the exact flag to fix.

## What changed

`validateNimMainPrefix` (new compile-time proc in `ffi/internal/ffi_library.nim`, called from `declareLibraryBase`) reads the compiler command line via `querySetting(SingleValueSetting.commandLine)` and:

- **Errors** when `--nimMainPrefix:X` is present but `X != "lib" & libraryName`, naming the expected flag (e.g. `needs --nimMainPrefix:libmy_timer, but the command line passes --nimMainPrefix:libtimer`).
- **Hints** (never errors) when the flag is absent — config.nims may set it without it surfacing on `commandLine`, so absence is not proof of a mistake. The hint is gated on `compileOption("app", "lib")` so it only speaks for real shared-library builds and stays silent for the unit tests, which stub `NimMain` via `{.emit.}` and legitimately never pass the flag.

`nimMainPrefixOnCmdLine` parses the flag tolerant of Nim's style-insensitive switch syntax (case-insensitive name, underscores stripped, `:` or `=` separator, last-one-wins).

Also fixes a latent mismatch the new check surfaces: `genbindings_cddl` in `ffi.nimble` passed `--nimMainPrefix:libtimer` for `declareLibrary("my_timer", ...)`; corrected to `libmy_timer` to match every other timer task.

## Tests

`tests/unit/test_main_prefix_validation.nim` drives a real `nim check` of `mainprefix_fixture.nim` (via `staticExec` at the test's own compile time) with a wrong prefix and the correct one, asserting the wrong build errors and names the expected flag, and the correct build is error-free. The fixture deliberately lacks the `test_` prefix so the nimble runner never compiles it standalone.

Verified: wrong prefix → compile error (replacing the link error), correct prefix builds `examples/timer` cleanly, hint silent for unit tests and live for the `--app:lib` timer build, and the new test passes under both `--mm:orc` and `--mm:refc`.

### DIFF
--- a/ffi.nimble
+++ b/ffi.nimble
@@ -226,7 +226,7 @@ task genbindings_rust, "Generate Rust bindings for the timer example":
     " -o:/dev/null examples/timer/timer.nim"
 
 task genbindings_cddl, "Generate CDDL schema for the timer example":
-  exec "nim c " & nimFlagsOrc & " --app:lib --noMain --nimMainPrefix:libtimer" &
+  exec "nim c " & nimFlagsOrc & " --app:lib --noMain --nimMainPrefix:libmy_timer" &
     " -d:ffiGenBindings -d:targetLang=cddl" &
     " -d:ffiOutputDir=examples/timer/cddl_bindings" & " -d:ffiSrcPath=../timer.nim" &
     " -o:/dev/null examples/timer/timer.nim"
--- a/ffi/internal/ffi_library.nim
+++ b/ffi/internal/ffi_library.nim
@@ -1,9 +1,58 @@
-import std/[macros, atomics, sysatomics], strformat, chronicles, chronos
+import
+  std/[macros, atomics, sysatomics, compilesettings], strformat, chronicles, chronos
+import strutils
 import ../codegen/meta
+
+func nimMainPrefixOnCmdLine(cmdLine: string): tuple[found: bool, value: string] =
+  ## Scan the compiler command line for `--nimMainPrefix:X` (last one wins) and
+  ## return its value. Switch names in Nim are style-insensitive, so the name
+  ## is matched lowercased and with underscores stripped; the separator may be
+  ## `:` or `=`. Returns `(false, "")` when the flag is absent — note config.nims
+  ## switches may not surface here, so absence is not proof it was never set.
+  var found = false
+  var value = ""
+  for tok in cmdLine.splitWhitespace():
+    let body = tok.strip(trailing = false, chars = {'-'})
+    let sep = body.find({':', '='})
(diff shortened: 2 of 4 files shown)


ffi/internal/ffi_library.nim
```diff
@@ -1,10 +1,59 @@
...
+    error(
+      "declareLibrary(\"" & libraryName &
+        "\"): the Nim runtime init symbol is importc'd as " & expectedPrefix &
+        "NimMain, so the build needs --nimMainPrefix:" & expectedPrefix &
+        ", but the command line passes --nimMainPrefix:" & prefixValue &
+        ". Change the flag to --nimMainPrefix:" & expectedPrefix &
+        " (it must be \"lib\" followed by the declareLibrary name)."
+    )
+  elif not prefixFound and compileOption("app", "lib"):
+    hint(
```
Ivansete-status, line 44: Should we enforce compile-time error too?

ffi.nimble
```diff
@@ -217,7 +217,7 @@ task genbindings_rust, "Generate Rust bindings for the timer example":
     " -o:/dev/null examples/timer/timer.nim"
 
 task genbindings_cddl, "Generate CDDL schema for the timer example":
-  exec "nim c " & nimFlagsOrc & " --app:lib --noMain --nimMainPrefix:libtimer" &
+  exec "nim c " & nimFlagsOrc & " --app:lib --noMain --nimMainPrefix:libmy_timer" &
```
Ivansete-status, line 229: 🫠

Ivansete-status (approved):

ffi/internal/ffi_library.nim
gmelodie: Good question! This is what our AI overlord told me

> A prefix set in config.nims does not appear in querySetting(commandLine). I tested it — with switch("nimMainPrefix", ...) in config.nims and no CLI flag, commandLine comes back with the flag absent. So "not found on the command line" ≠ "not set".
> 
> Also, nim exposes no accessor for the effective (config-resolved) prefix. There's no SingleValueSetting for it and no compileOption equivalent. The command-line scan is the only signal available at compile time.

PR #119, opened 2026-07-09
feat: shrink the genbindings incantation

gmelodie:
## Summary

Binding generation used to need seven flags repeated per language, per file: `--app:lib --noMain --nimMainPrefix:… -d:ffiGenBindings -d:targetLang=… -d:ffiOutputDir=… -d:ffiSrcPath=… -o:/dev/null`. `ffi.nimble` carried ~10 near-identical tasks, `examples/timer/timer.nimble` replicated them, and every downstream library would have inherited the same boilerplate. This collapses the incantation down to what actually matters.

## What changed

- **One compile emits every language.** `-d:targetLang` now accepts a comma-separated list (`rust,cpp,c,cddl`) and `genBindings()` loops the backend dispatch over each token, so a single `nim c` produces all of them instead of one invocation per language.
- **Output dir is derived, not spelled out.** It defaults to `<lang>_bindings/` next to the compiled source via `querySetting(projectPath)`; `-d:ffiOutputDir` still overrides.
- **Source path is derived, not spelled out.** The `ffiSrcPath` embedded in generated build files is computed with `relativePath` from the output dir to `querySetting(projectFull)`; `-d:ffiSrcPath` still overrides. This also fixes a latent inconsistency where `timer.nimble` embedded `timer.nim` while the checked-in files use `../timer.nim`.
- **`--compileOnly` replaces `-o:/dev/null`.** Macros expand during sem, so the binding files are written before any linking — `--compileOnly` is enough and the `--app:lib --noMain --nimMainPrefix` flags are dropped from the generation pass (the real link flags live in `ffi/codegen/templates/nim_ffi_lib.cmake`, independent of how bindings are generated).
- **Nimble tasks ride a small `genBindingsCmd` helper.** Nimble files can't import package modules, so each keeps a short local proc; the ~10 timer/echo tasks are now one-liners.
- The per-language `case` dispatch in `genBindings()` is extracted into `emitBindingsFor(...)` to keep the macro body flat.

## Testing

- `nimble check_bindings` passes under both `--mm:orc` and `--mm:refc` — the regenerated Rust, C++, C, abi=c C and CDDL bindings are byte-identical to the checked-in files (the risk called out for this change: path derivation must produce identical `ffiSrcPath` strings).
- Verified the dropped `--app:lib --noMain --nimMainPrefix` flags produce byte-identical output for every target, so the shorter generation pass is provably equivalent.
- `ffi.nim` still builds without `-d:ffiGenBindings` — the new `std/os` / `std/compilesettings` imports and helpers don't affect the normal library build.

### DIFF
--- a/examples/timer/timer.nimble
+++ b/examples/timer/timer.nimble
@@ -12,20 +12,24 @@ requires "https://github.com/logos-messaging/nim-ffi >= 0.2.0"
 
 const nimFlags = "--mm:orc -d:chronicles_log_level=WARN"
 
+proc genBindingsCmd(langs: string): string =
+  ## One `nim c` that emits `langs` (comma-separated) from timer.nim, each into
+  ## `<lang>_bindings/`. `--compileOnly` is enough — the files are written during
+  ## macro expansion, nothing is linked.
+  "nim c " & nimFlags & " -d:ffiGenBindings -d:targetLang=" & langs &
+    " --compileOnly timer.nim"
+
 task build, "Compile the timer library":
   exec "nim c " & nimFlags & " --app:lib --noMain --nimMainPrefix:libmy_timer timer.nim"
 
+task genbindings, "Generate Rust, C++ and C bindings for the timer example":
+  exec genBindingsCmd("rust,cpp,c")
+
 task genbindings_rust, "Generate Rust bindings for the timer example":
-  exec "nim c " & nimFlags & " --app:lib --noMain --nimMainPrefix:libmy_timer" &
-    " -d:ffiGenBindings -d:targetLang=rust" & " -d:ffiOutputDir=rust_bindings" &
-    " -d:ffiSrcPath=timer.nim" & " -o:/dev/null timer.nim"
+  exec genBindingsCmd("rust")
 
 task genbindings_cpp, "Generate C++ bindings for the timer example":
-  exec "nim c " & nimFlags & " --app:lib --noMain --nimMainPrefix:libmy_timer" &
-    " -d:ffiGenBindings -d:targetLang=cpp" & " -d:ffiOutputDir=cpp_bindings" &
-    " -d:ffiSrcPath=timer.nim" & " -o:/dev/null timer.nim"
+  exec genBindingsCmd("cpp")
 
 task genbindings_c, "Generate C bindings for the timer example":
-  exec "nim c " & nimFlags & " --app:lib --noMain --nimMainPrefix:libmy_timer" &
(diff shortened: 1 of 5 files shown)


examples/timer/timer.nimble
```diff
@@ -12,20 +12,24 @@ requires "https://github.com/logos-messaging/nim-ffi >= 0.2.0"
 
 const nimFlags = "--mm:orc -d:chronicles_log_level=WARN"
 
+proc genBindingsCmd(langs: string): string =
```
Ivansete-status, line 15: That is superb!

ffi/internal/ffi_macro.nim
```diff
@@ -1841,6 +1843,55 @@ macro ffiEvent*(args: varargs[untyped]): untyped =
...
   return generated
 
+proc bindingsOutputDir(lang, explicit: string): string {.compileTime.} =
+  ## Output dir for `lang`. Defaults to `<lang>_bindings/` next to the source
+  ## file being compiled (`querySetting(projectPath)`); an explicit
+  ## -d:ffiOutputDir override wins.
+  if explicit.len > 0:
+    explicit
+  else:
+    querySetting(SingleValueSetting.projectPath) / (lang & "_bindings")
```
Ivansete-status: I prefer being always explicit so that I don't have to lookup and check if this is returning something or is just a "void" call. Applies elsewhere.
```suggestion
    return querySetting(SingleValueSetting.projectPath) / (lang & "_bindings")
```

ffi/internal/ffi_macro.nim
```diff
@@ -1841,6 +1843,55 @@ macro ffiEvent*(args: varargs[untyped]): untyped =
...
+        genProcs, ffiTypeRegistry, libName, outDir, srcRel, ffiEventRegistry
+      )
+    of "c":
+      generateCBindings(
+        genProcs, ffiTypeRegistry, libName, outDir, srcRel, ffiEventRegistry
+      )
+    of "c_abi":
+      generateCAbiBindings(
+        genProcs, ffiTypeRegistry, libName, outDir, srcRel, ffiEventRegistry
+      )
```
Ivansete-status, line 1912: I mentioned in an ongoing PR but, what is the diff between these two?

Ivansete-status (approved):
Wonderful simplification, thanks!

ffi/internal/ffi_macro.nim
gmelodie: Ah! So, `c_abi` doesn't go through CBOR for functions that are scalar-only (scalar fast path), while `c` uses CBOR as an intermediate for everything

PR #120, opened 2026-07-09
feat(ffi): make the scalar-fast-path binding drop loud

gmelodie:
## Summary

An all-scalar `abi = c` proc dispatches through the CBOR-free scalar fast path, and the foreign-binding generators have no codegen for that shape yet — so under `-d:ffiGenBindings` such a proc was **silently dropped** from the generated bindings behind nothing louder than a compile hint. The consumer only found out when their own build failed to link a function that should have been in the header.

This makes the drop loud: `genBindings()` now **fails** and names the affected procs. Full scalar-shape foreign codegen is the real fix (L-sized, tracked as follow-up); erroring here makes the gap safe today instead of shipping a header with a hole in it.

## Behavior

Under `-d:ffiGenBindings`, if any proc rides the scalar fast path, `genBindings()` errors with the offending proc names and three remedies: switch it to `abi = cbor`, add a non-scalar param so it takes the CBOR wire shape, or pass `-d:ffiAllowScalarSkip` to accept the omission (which downgrades the error back to a per-proc hint, the old behavior).

Without `-d:ffiGenBindings`, `genBindings()` is a no-op as before — nothing changes for a normal library build.

## Changes

- `ffi/internal/ffi_macro.nim` — new `reportScalarFastPathDrops` proc (guard-claused) that collects scalar-fast-path procs and either errors or hints; called from `genBindings()` in place of the old inline per-proc `hint`.
- `ffi/codegen/meta.nim` — `-d:ffiAllowScalarSkip` (`booldefine`, default off) — the escape hatch.
- `ffi.nimble` — the `genbindings_c_abi_echo` task now passes `-d:ffiAllowScalarSkip`: `echoVersion` is all-scalar under the echo lib's `abi = c` default and was already being omitted from `examples/echo/c_abi_bindings`, so the flag preserves that intended omission (generated headers stay byte-identical).
- `README.md` — ABI section documents the fast path, the error, and the escape hatch.

## Testing

- `tests/unit/test_scalar_skip_gen.nim` + `tests/unit/fixtures/scalar_skip_fixture.nim` (new) — compiles a fixture library with an all-scalar `abi = c` proc in a child `nim check` and asserts (1) generation fails naming the dropped proc with the remedy text, and (2) `-d:ffiAllowScalarSkip` builds clean with no `Error:`. Green under `--mm:orc`.
- Verified no regression: `test_scalar_fastpath` and `test_c_abi_codegen` still pass; `nimble genbindings_c_abi_echo` succeeds with `examples/echo/c_abi_bindings` unchanged; the library still builds `--app:lib`.

## Follow-up

Real fix is full foreign-binding codegen for the scalar shape (L-sized) — until then, any downstream library with an all-scalar `abi = c` proc must add `-d:ffiAllowScalarSkip` or adjust the proc.

### DIFF
--- a/ffi/codegen/meta.nim
+++ b/ffi/codegen/meta.nim
@@ -173,7 +173,7 @@ const ffiOutputDir* {.strdefine.} = ""
 # set with -d:ffiSrcPath=../relative/path.nim
 const ffiSrcPath* {.strdefine.} = ""
 
-# When set, scalar-only `abi = c` procs (which have no foreign-binding codegen
+# When set to true, scalar-only `abi = c` procs (which have no foreign-binding codegen
 # yet) are silently omitted from the generated bindings instead of failing the
 # build. Off by default so the drop is loud; see genBindings().
 const ffiAllowScalarSkip* {.booldefine.} = false
--- a/ffi/codegen/meta.nim
+++ b/ffi/codegen/meta.nim
@@ -172,3 +172,8 @@ const ffiOutputDir* {.strdefine.} = ""
 # Nim source path (relative to outputDir) embedded in generated build files;
 # set with -d:ffiSrcPath=../relative/path.nim
 const ffiSrcPath* {.strdefine.} = ""
+
+# When set, scalar-only `abi = c` procs (which have no foreign-binding codegen
+# yet) are silently omitted from the generated bindings instead of failing the
+# build. Off by default so the drop is loud; see genBindings().
+const ffiAllowScalarSkip* {.booldefine.} = false
--- a/ffi/internal/ffi_macro.nim
+++ b/ffi/internal/ffi_macro.nim
@@ -1841,6 +1841,32 @@ macro ffiEvent*(args: varargs[untyped]): untyped =
     echo generated.repr
   return generated
 
+proc reportScalarFastPathDrops(procs: seq[FFIProcMeta]) {.compileTime.} =
+  ## Scalar-fast-path procs have no foreign-binding codegen yet, so they can't
+  ## ride the generated bindings. Fail loudly, naming them, unless
+  ## `-d:ffiAllowScalarSkip` opts into the silent omission (then just hint).
+  var skipped: seq[string] = @[]
+  for p in procs:
+    if p.scalarFastPath:
+      skipped.add(p.procName)
+  if skipped.len == 0:
+    return
+  if ffiAllowScalarSkip:
+    for name in skipped:
+      hint(
+        "genBindings: omitting scalar-fast-path proc '" & name &
+          "' from the bindings (-d:ffiAllowScalarSkip)"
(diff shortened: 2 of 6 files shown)


ffi/codegen/meta.nim
```diff
@@ -172,3 +172,8 @@ const ffiOutputDir* {.strdefine.} = ""
 # Nim source path (relative to outputDir) embedded in generated build files;
 # set with -d:ffiSrcPath=../relative/path.nim
 const ffiSrcPath* {.strdefine.} = ""
+
+# When set, scalar-only `abi = c` procs (which have no foreign-binding codegen
```
Ivansete-status: Maybe?
```suggestion
# When set to true, scalar-only `abi = c` procs (which have no foreign-binding codegen
```

---

What is the result of a proc being silently omitted? It sounds to me that it should never be skipped at all. Or does it mean that they are automatically CBOR-generated?

ffi/internal/ffi_macro.nim
```diff
@@ -1841,6 +1841,32 @@ macro ffiEvent*(args: varargs[untyped]): untyped =
     echo generated.repr
   return generated
 
+proc reportScalarFastPathDrops(procs: seq[FFIProcMeta]) {.compileTime.} =
+  ## Scalar-fast-path procs have no foreign-binding codegen yet, so they can't
```
Ivansete-status, line 1845: Ok, I see that next natural step is to tackle the generation problem, at least for C++?

tests/unit/fixtures/scalar_skip_fixture.nim
```diff
@@ -0,0 +1,26 @@
...
+type SkipConfig {.ffi.} = object
+  base: int
+
+proc scalarskip_create*(cfg: SkipConfig): Future[Result[SkipLib, string]] {.ffiCtor.} =
+  return ok(SkipLib(base: cfg.base))
+
+proc scalarskip_add*(
+    lib: SkipLib, a: int, b: int
+): Future[Result[int, string]] {.ffi: "abi = c".} =
+  ## All-scalar signature: dispatches through the CBOR-free fast path and has no
```
Ivansete-status, line 22: Ah okay, then, on _scalar_ procs they now use cbor ABI regardless the pragma value.

Ivansete-status (approved):

ffi/internal/ffi_macro.nim
gmelodie: Yes! absolutely

tests/unit/fixtures/scalar_skip_fixture.nim
gmelodie: I think this was already the case (implemented in https://github.com/logos-messaging/nim-ffi/pull/110)

PR #121, opened 2026-07-09
docs: complete the README workflow story

gmelodie:
## Summary

The README documented the pragmas and the minimal example but left out the parts a downstream author actually trips over when going from "annotated procs" to "shipped, bound library." This fills those gaps — docs-only, no code changes.

Added / expanded in `README.md`:

- **Install line** — a new `## Install` section with the `requires "https://github.com/logos-messaging/nim-ffi >= 0.2.0"` line downstreams need (matching what `examples/timer` and `examples/echo` already use).
- **Two-compile model** — rewrote `## Generating bindings` into `## Building — the two-compile model`, explaining *why* a library ships from two compiles of the same source: the normal build produces just the shared library, and binding emission is gated behind `-d:ffiGenBindings` (so `genBindings()` is a no-op otherwise). The binding compile runs the generators as a compile-time side effect and produces no runnable output, hence `-o:/dev/null`. Both commands are shown; `c_abi` added to the `-d:targetLang` list.
- **The `--nimMainPrefix:lib<name>` rule** — new subsection explaining `--app:lib` / `--noMain` and why the prefix must be exactly `lib` + the `declareLibrary` name on both compiles: `declareLibrary` emits `initializeLibrary()` which calls `lib<name>NimMain()`, so a mismatch fails to link.
- **Library-naming collision note** — distinct name/prefix per library when several coexist in one process (the C++ e2e loads `libmy_timer` + `libecho`).
- **Honest `abi = c` limits** — expanded `### ABI format`: events under `abi = c` are a hard compile error (must be `abi = cbor`), and all-scalar `abi = c` procs take the CBOR-free scalar fast path at runtime but are dropped from the foreign bindings (codegen is a follow-up).
- **Request timeouts** — new `### Request timeouts` subsection documenting the 5s `DefaultRequestTimeout` (`ffi/ffi_context.nim`), the exact `ffi request timed out after <n>ms` trip behavior (caller unblocked, handler left running, callback fires once), and the `{.ffi: "timeout = <ms>".}` override — previously only in the CHANGELOG.

All facts verified against source (`ffi/ffi_context.nim`, `ffi/ffi_thread.nim`, `ffi/internal/ffi_macro.nim`, `ffi/internal/ffi_scalar.nim`, `ffi/internal/ffi_library.nim`) and the example `.nimble` build tasks.

## Testing

Docs-only change — no build or test surface. Verified error strings, define names, default values, and the `--nimMainPrefix` derivation against the referenced source files.


README.md
```diff
@@ -111,6 +140,24 @@ The default wire format is `cbor`. Override the library default with
 `declareLibrary("lib", Lib, defaultABIFormat = "c")`, or per annotation with an
 `"abi = ..."` spec, e.g. `{.ffi: "abi = c".}`.
 
+`cbor` is the fully-supported format: every proc, ctor, dtor and event
+serializes through the generic CBOR path, and all four binding backends emit
+working callers for it.
+
+`abi = c` (flat C-struct wire, generated by `-d:targetLang=c_abi`) is newer and
```
Ivansete-status: I thought that the target ABI can only be changed from within the code.
1. At declare time: `declareLibrary("mylib", MyLib, defaultABIFormat = "c")`
2. Override 1. on each proc with either `.ffi: "abi = c".` or `.ffi: "abi ) cbor".`.

Besides, the `targetLang` compile flag sounds related to the auto-generated code and I feel confused when seeing `c_abi` there because that is not a expected language (c, cpp, rust, etc.)

README.md
```diff
@@ -111,6 +140,24 @@ The default wire format is `cbor`. Override the library default with
...
+carries two honest limits today:
+
+- **Events are CBOR-only.** Applying `abi = c` to an `{.ffiEvent.}` proc is a
+  hard compile error; declare events with `abi = cbor` (they ride CBOR
+  internally regardless of the library default).
+- **All-scalar `abi = c` procs are dropped from the foreign bindings.** A
+  `{.ffi: "abi = c".}` method whose every param and return is a plain scalar
+  takes the CBOR-free scalar fast path at runtime, but the foreign codegen for
+  that inline-args shape is a follow-up — such procs are omitted from the
```
Ivansete-status: Do we have an issue for that follow-up?

README.md
```diff
@@ -122,20 +169,50 @@ once at the bottom of the top-level root file.
...
+## Building — the two-compile model
+
+A nim-ffi library ships from **two separate compiles of the same source**,
+because binding emission is gated behind `-d:ffiGenBindings`: without that
+define `genBindings()` is a no-op, so the normal build just produces the shared
+library and nothing else.
 
-Binding emission is gated behind `-d:ffiGenBindings`; without it, `genBindings()`
-is a no-op and the library just builds normally.
+**1. Build the shared library** (the artifact your host loads):
```
Ivansete-status, line 182: For this point, I think is interesting to mention that it generated the .h and we need to explain where the .h is put.

README.md
```diff
@@ -122,20 +169,50 @@ once at the bottom of the top-level root file.
...
+runnable output, so send the binary to `/dev/null`:
+
+```sh
+nim c --app:lib --noMain --nimMainPrefix:libmylib \
+      -d:ffiGenBindings -d:targetLang=c \
+      -d:ffiOutputDir=path/to/output -d:ffiSrcPath=../mylib.nim \
+      -o:/dev/null mylib.nim
+```
+
+- `-d:targetLang` — `rust` (default), `cpp`, `c`, `c_abi`, or `cddl`.
```
Ivansete-status: We may need to reconsider the `c_abi` and `cddl` as they don't seem belonging to the same group as `rust`, `cpp`, or `c`.

Ivansete-status (approved):
LGTM! Thanks for it! 🙌 
I noticed that there are some points that we may need to revisit in further PRs.

README.md
gmelodie: Agreed! https://github.com/logos-messaging/nim-ffi/issues/125

README.md
gmelodie: Better! We have an open PR!! https://github.com/logos-messaging/nim-ffi/pull/124

PR #124, opened 2026-07-09
feat(codegen): scalar-fast-path bindings for the abi = c header

gmelodie:
## Summary

Implements foreign-binding codegen for scalar-fast-path `abi = c` procs — the follow-up that #120 tracked with a loud placeholder drop.

An all-scalar `abi = c` proc (fixed-width int/float/bool params, an optional `string` return, no structs/handles/pointers) dispatches through the CBOR-free scalar fast path from #106/#110, but had no foreign binding. Under `-d:ffiGenBindings` it was dropped from every generated header, and since #120 that drop fails the build unless `-d:ffiAllowScalarSkip` is passed — so the fast path was only reachable from hand-written callers.

Since #130 there is a single `c` target whose header shape (CBOR vs `abi = c`) is chosen from the library's ABI format. This PR teaches the `abi = c` C header to bind the scalar shape directly, leaving the dylib ABI unchanged:

- The extern decl takes the scalar args inline plus a raw `FFICallBack`-shaped callback — no request struct crosses: `int echo_version(void* ctx, EchoScalarRawFn callback, void* user_data);`.
- A per-method heap box + trampoline adapt the raw-bytes reply into the same typed `ReplyFn` surface the flat-struct methods expose: a `string` return rides as raw UTF-8 (copied + NUL-terminated, since wire payloads aren't NUL-terminated); every other scalar is unpacked from the 8-byte native-endian scalar image (sign-extended ints narrowed back, `float32` re-narrowed from the widened double, bool from 0/1). Error messages are copied + NUL-terminated the same way.
- The reply/error copies route through a CBOR-free `nimffi_abi_dup_cstr_n` helper emitted once per header (include-guarded so two `abi = c` headers can co-exist in one TU) that rejects a length which would overflow `malloc(n + 1)`; the shared CBOR-side `nimffi_dup_cstr_n` gained the same `SIZE_MAX` guard.
- The box is freed by the trampoline, which the dylib invokes exactly once on every path (ctx-guard and enqueue failures reply synchronously; success replies from the FFI thread) — the same contract the flat-struct reply trampolines rely on.

`genBindings()` binds scalar procs only for the `c` target of an `abi = c` library (`currentDefaultABIFormat == ABIFormat.C`). Every other target — and the CBOR C header — still drops scalar procs loudly, unless `-d:ffiAllowScalarSkip`.

## Affected areas

- `ffi/codegen/c.nim`: scalar extern decls, the per-method reply box + trampoline, and the high-level `lib_ctx_<name>` wrappers; the Req envelope is skipped for scalar procs.
- `ffi/internal/ffi_macro.nim`: `genBindings()` routes the full registry to the `abi = c` C header and drops+reports for everything else; honest drop-error text.
- `ffi/codegen/meta.nim`, `ffi/internal/ffi_scalar.nim`: doc comments aligned to the single-`c`-target model.
- `ffi.nimble`: `genbindings_c_abi_echo` no longer needs `-d:ffiAllowScalarSkip`.
- `ffi/codegen/templates/c/cbor_helpers.h.tpl`: `nimffi_dup_cstr_n` now returns NULL on a `SIZE_MAX` length instead of wrapping `malloc(n + 1)` to zero.
- `examples/echo/c_abi_bindings/echo.h`, `examples/echo/c_bindings/nim_ffi_cbor.h`, `examples/timer/c_bindings/nim_ffi_cbor.h`: regenerated; the abi=c header now carries `echo_ctx_version` and the guarded dup helper.
- Tests: `test_c_abi_codegen` gains cases for the scalar extern shape, the typed reply typedefs, the box/trampoline/wrapper, and the int/float32/string unpack shapes; `test_scalar_skip_gen` covers the loud drop and the `-d:ffiAllowScalarSkip` downgrade; the c_abi e2e drives `echo_ctx_version` against the real dylib.

## Impact on users

- `abi = c` consumers gain bindings for scalar-fast-path procs; purely additive — existing flat-struct bindings and exported dylib symbols are unchanged.
- Builds that passed `-d:ffiAllowScalarSkip` under an `abi = c` library no longer need it; the flag keeps its meaning for the CBOR outputs.

## Verification

- `test_c_abi_codegen` 12/12, `test_scalar_skip_gen` 2/2.
- The `abi = c` `echo.h` regenerates cleanly with `-d:targetLang=c` and no skip flag; the regenerated headers compile under `gcc`/`g++` with `-Wall -Wextra` (including a double-include of the abi=c header).

## References

- #106, #110 — CBOR-free scalar fast path (runtime dispatch).
- #109 — original `abi = c` C bindings; #130 unified the C and c_abi generators into one `c` target that picks the header shape from the library ABI format.
- #120 — made the scalar drop loud as a placeholder; this PR is the binding codegen it was waiting for.


ffi/codegen/c_abi.nim
```diff
@@ -131,7 +131,9 @@ proc newAbiReg(types: seq[FFITypeMeta], procs: seq[FFIProcMeta]): AbiReg =
   for t in types:
     reg.typeTable[t.name] = t
   for p in procs:
-    if p.kind != FFIKind.DTOR:
+    # A scalar-fast-path proc has no Req envelope: its export takes the scalar
+    # args inline (see emitScalarMethod).
+    if p.kind != FFIKind.DTOR and not p.scalarFastPath:
```
Ivansete-status: Does "fast path" mean "args inline" ? It might be explained somewhere but "fast path" terms is a little bit wide concept.
```suggestion
    if p.kind != FFIKind.DTOR and not p.scalarArgsInline:
```

examples/echo/c_bindings/nim_ffi_cbor.h
```diff
@@ -271,8 +271,12 @@ static inline char* nimffi_dup_cstr(const char* s) {
 }
 
 /* NUL-terminated copy of a length-delimited (not NUL-terminated) byte run,
- * for turning the FFICallback's raw error `msg`/`len` into a C string. */
+ * for turning the FFICallback's raw error `msg`/`len` into a C string. Returns
+ * NULL on allocation failure or a length that would overflow `n + 1`. */
 static inline char* nimffi_dup_cstr_n(const char* s, size_t n) {
+    if (n == SIZE_MAX) {
```
Ivansete-status, line 277: The suggestion applies to other files too.
```suggestion
    if (n > SIZE_MAX) {
```

ffi/internal/ffi_macro.nim
```diff
@@ -1535,10 +1537,13 @@ proc reportScalarFastPathDrops(procs: seq[FFIProcMeta]) {.compileTime.} =
       )
     return
   error(
-    "genBindings: no foreign-binding codegen for scalar-fast-path `abi = c` " &
-      "procs yet, so these would be silently omitted from the generated " & "bindings: " &
-      skipped.join(", ") & ".\n" & "Fix by one of:\n" &
-      "  - switch the proc to `abi = cbor`, or\n" &
+    "genBindings: this target has no foreign-binding codegen for " &
```
Ivansete-status: We can make this snippet a bit more readable with `"""`.
E.g.:
```nim
    """genBindings: this target has no foreign-binding codegen for scalar-fast-path
`abi = c` procs, so these would be silently omitted from the generated bindings:
$1
They are emitted only into the `abi = c` C header (an `abi = c` library generated
with -d:targetLang=c).
Fix by one of:
  - make the library `abi = c` (declareLibrary(..., "c")) and generate C bindings, or
  - switch the proc to `abi = cbor`, or
  - add a non-scalar param (e.g. a struct or handle) so it takes the CBOR wire shape, or
  - pass -d:ffiAllowScalarSkip to accept the omission.""" % [skipped.join(", ")]
```

ffi/internal/ffi_scalar.nim
```diff
@@ -32,8 +32,11 @@ func isScalarOnly*(p: FFIProcMeta): bool =
   true
 
 func bindableProcs*(procs: seq[FFIProcMeta]): seq[FFIProcMeta] =
-  ## Procs the foreign-binding generators emit for; scalar-fast-path procs are
-  ## dropped (their inline-scalar export doesn't match the CBOR codegen shape).
+  ## The procs the CBOR-speaking foreign-binding generators emit for.
```
Ivansete-status: Maybe too long comment. Applies elsewhere.

README.md
```diff
@@ -171,13 +171,18 @@ header shape from the library's ABI format. It carries two honest limits today:
...
+(no request struct) and adapts the raw-bytes reply into the same typed callback
+surface the flat-struct methods use. The CBOR-speaking outputs — the CBOR C
+header and the `cpp`, `rust`, `cddl` targets — have no scalar codegen, so under
+`-d:ffiGenBindings` they would omit such a proc from the generated bindings —
+and `genBindings()` fails with an error naming the affected procs. Resolve it by
+making the whole library `abi = c` and generating C bindings, switching the proc
+to `abi = cbor`, adding a non-scalar param so it takes the CBOR wire shape, or
+passing `-d:ffiAllowScalarSkip` to accept the omission (the proc still works
+over the scalar fast path; it's just absent from the generated foreign
+bindings).
```
Ivansete-status: If possible, let's make it more concise.
```suggestion
**All-scalar `abi = c` procs bind only in the `abi = c` C header.** A
  `{.ffi: "abi = c".}` method whose params and return are all scalars — ints,
  floats, bools; a `string` return is fine, a `string` param is not — takes a
  CBOR-free fast path, and its C wrapper passes the args inline instead of
  packing a request struct. Only the `abi = c` C header emits that shape. The
  CBOR C header and the `cpp`, `rust` and `cddl` targets have no scalar codegen
  and would silently omit the proc, so `genBindings()` fails and names it. Fix
  it by generating C bindings from an `abi = c` library, switching the proc to
  `abi = cbor`, giving it a non-scalar param, or passing
  `-d:ffiAllowScalarSkip` to accept the omission — the proc still works over the
  fast path, it's just absent from the bindings.
```

Ivansete-status (approved):
LGTM! Thanks for it! 🙌

PR #128, opened 2026-07-10
fix(test): rebuild CBOR libecho in cpp e2e to avoid ABI mismatch

Ivansete-status:
## Problem

`TimerE2E.CrossLibrary` (the only e2e that loads two nim-ffi dylibs in one process) **segfaults** when the C++ CBOR suite runs after the `abi=c` suite in the same workspace.

Backtrace: `libecho.dylib`echo_create → cwireUnpack → cstrToNimstr → strlen` on a garbage `char*` (the faulting address literally contains the ASCII bytes of "config").

## Root cause — build hygiene, not a runtime bug

The CBOR and `abi=c` echo suites compile `examples/echo/echo.nim` to the **same** `${REPO_ROOT}/libecho.dylib` (the name derives only from the lib name, not the ABI). CMake keys the rebuild on `echo.nim`'s mtime, which does **not** change when only the `-d:ffiEchoAbiC` flag flips. So flipping the ABI silently reuses the previously built dylib.

`removeStaleEchoLib()` guarded only the `abi=c` tasks. Running an `abi=c` suite first left the flat-struct dylib in place; the CBOR `CrossLibrary` test then reached `echo_create`'s `cwireUnpack`, decoded CBOR bytes as a flat struct, and crashed. (`removeStaleEchoLib`'s own comment described only the reverse direction.)

## Fix

Call `removeStaleEchoLib()` in the CBOR cpp e2e tasks too (`test_cpp_e2e`, `test_cpp_e2e_sanitized`), and update the helper's doc to note the hazard is bidirectional.

## Verification

Repro'd the exact bad sequence, then confirmed the fix:

```
nimble test_c_abi_e2e   # writes abi=c libecho.dylib
nimble test_cpp_e2e     # before: CrossLibrary SEGFAULT; after: 20/20 pass
```

### DIFF
--- a/ffi.nimble
+++ b/ffi.nimble
@@ -88,10 +88,11 @@ proc removeStaleEchoLib() =
   ## The CBOR and `abi = c` echo e2e suites both compile examples/echo/echo.nim
   ## to the same repo-root `libecho.so`, differing only by `-d:ffiEchoAbiC`.
   ## CMake keys the dylib rebuild on echo.nim's mtime, not the ABI flag, so a
-  ## `libecho.so` left by an earlier CBOR e2e step is silently reused by the
-  ## abi=c build — the flat-struct C caller then reaches the CBOR entry points
-  ## (wrong arity/ABI) and segfaults. Delete it first to force a fresh rebuild
-  ## with the right ABI.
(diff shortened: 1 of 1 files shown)


gmelodie (approved):
lgtm!

PR #129, opened 2026-07-10
feat(ffi): RET_STALE_WARN progress callback replacing handler timeout

### LINKED ISSUE #126: Review per-function timeout design and document its scope
Reviewing the per-function timeout currently baked into nim-ffi.

Open question:
- Should nim-ffi set this timeout itself, or should it be left to the caller to configure? Hardcoding a timeout on our side removes control from the user; exposing it (or letting them pass one in) may be the better default.

Behavior to document regardless of the above:
- The 30s timeout applies to the *whole operation*, not a single connection attempt.
- A dial can be given multiple addresses, and nim-libp2p dials them sequentially. So the 30s budget is shared across all address attempts, not spent per-address.
- This needs to be called out clearly in the exported API docs so callers understand that passing N addresses does not give them N * 30s.

Decide on ownership of the timeout (us vs. caller), then document the final semantics.

https://github.com/vacp2p/nim-libp2p/pull/2774#discussion_r3558827207

Ivansete-status:
Closes #126.

## What

nim-ffi **never times a handler out** — a hard-cancel mid-call into the underlying library (Waku/libp2p) can leave it half-applied. Instead, while a request is still in flight its result callback receives a **non-terminal `RET_STALE_WARN` (3)** every `StaleWarnInterval` (5s default; `-d:ffiStaleWarnIntervalMs=<ms>` to override), with the payload carrying the elapsed milliseconds as a decimal string. The request **always** ends with exactly one terminal `RET_OK` / `RET_ERR`.

The dev decides what to do with a slow request — keep waiting, show a spinner, tear the context down — nim-ffi does not decide for them.

## Why

This replaces the never-released per-proc `{.ffi: "timeout = <ms>".}` override and the `defaultRequestTimeout` context field (#93), whose hard abort unblocked the caller early. Per the discussion in #126, a fixed hard timeout can't know the right budget for a given call (e.g. sequential multi-address dials), so the caller is kept informed and owns the decision. The stale-warn interval is an internal context seam (`staleWarnInterval`), **not** a per-proc pragma.

## Callback contract

| Code | Value | Terminal? | Meaning |
| --- | --- | --- | --- |
| `RET_OK` | 0 | yes | Success; payload = encoded result |
| `RET_ERR` | 1 | yes | Failure; payload = UTF-8 error |
| `RET_MISSING_CALLBACK` | 2 | — | No callback passed |
| `RET_STALE_WARN` | 3 | **no** | Still running; payload = elapsed ms |

## Codegen / safety

- Emit `RET_STALE_WARN` in the C / C++ headers (documented).
- Every generated reply trampoline (C, C++, Rust, and the Nim `abi=c` reply trampolines) frees its one-shot box on the first callback — so each now **early-returns on the non-terminal code** to avoid a double-free from the repeated callback. Typed wrappers ignore the progress signal for now; it is delivered at the raw result-callback boundary.

## Tests

- New end-to-end test in `test_ffi_context.nim`: a slow handler pings the caller several times (invariant: the k-th ping reports exactly `k * interval`) then delivers one terminal `RET_OK`, with nothing trailing it.
- Dropped the #93 timeout suites.
- Full unit suite green (orc + refc); C / C++ / c_abi e2e green.

## Note

Depends on the build-hygiene fix in #128 for the full e2e matrix to run reliably in a shared workspace (independent PR).


gmelodie (approved):
Good! I'd just reduce the size of comments throughout if possible

PR #143, opened 2026-07-31
abi = c header: self-contained, event-listener ABI, doc comments, ctor/dtor guard

gmelodie:
## Summary

Rounds out the generated C bindings so an `abi = c` consumer needs no hand-written header, and adds two compile-time ergonomics to `declareLibrary` / `genBindings`.

The driver was issue #6 (a Waku kernel that hand-writes the `abi = c` header). Each change below removes one reason that header had to exist.

## Changes

- **Event-listener ABI in the `abi = c` header** (`ffi/codegen/c.nim`, `emitAbiExternDecls`). `declareLibrary` always exports `<lib>_add_event_listener` / `<lib>_remove_event_listener` on the `FFIContext`, but the `abi = c` generator never declared them. It now emits both, plus the `FFICallback` typedef they take (guarded by `NIMFFI_FFICALLBACK_DEFINED`, and spelled to match the CBOR/C++ headers). The typed `{.ffiEvent.}` listener machinery is still unsupported under `abi = c` — this is only the raw ABI.
- **Self-contained `abi = c` header** (`ffi/codegen/c.nim`). Beside the existing `<stdint.h>`/`<stddef.h>` includes and `NIMFFI_RET_*` codes, the header now emits short `#ifndef`-guarded `RET_*` aliases (`RET_OK`, `RET_ERR`, `RET_MISSING_CALLBACK`, `RET_STALE_WARN`) for consumers that use the unprefixed names.
- **`ABIFormat` enum overload of `declareLibrary`** (`ffi/internal/ffi_library.nim`). The shared body moved into `declareLibraryImpl`; a second macro overload accepts `defaultABIFormat: static[ABIFormat]`, so `defaultABIFormat = ABIFormat.C` compiles next to the `"c"` string. Two-arg calls stay unambiguous because the enum overload has no default for that parameter.
- **`headerBanner` option** (`declareLibrary` → `currentHeaderBanner` → `generateCBindings` → both C headers). An optional string stamped as `//`-comment lines at the top of every generated C header (e.g. a "generated — do not edit" note). A trailing backslash in a banner line is stripped so it cannot splice the include guard into the comment.
- **ctor-without-dtor is a compile error** (`ffi/internal/ffi_macro.nim`, `genBindings`, issue #3). A `{.ffiCtor.}` allocates a context; without a `{.ffiDtor.}` the caller cannot release the underlying Nim library object. `genBindings()` now fails, naming the ctor and the fix. The check fires at binding generation, so a runtime-only library that never calls `genBindings()` is not covered.

Doc-comment propagation into the header (issue #6's "USE AT YOUR OWN RISK" note) already landed earlier on this branch; `test_doc_propagation` covers it and it needs no change.

## Regenerated bindings

- `examples/echo/c_abi_bindings/echo.h` — regenerated under orc and refc; the diff is the new listener decls, `FFICallback` typedef and `RET_*` aliases.

## Tests

- New `tests/unit/test_declare_library.nim` — child `nim check` fixtures for the enum overload (compiles, sets the default) and the ctor-without-dtor error (fails, names the ctor).
- New suites in `tests/unit/test_c_codegen.nim` — the `abi = c` header's includes, `RET_*` aliases, event-listener ABI, `FFICallback` guard, the banner (both header shapes), and a regression that a backslash-terminated banner line cannot delete the include guard.
- Added the required `{.ffiDtor.}` to three fixtures that declared a ctor without one (`test_c_abi_dispatch`, `ffistatic_abi_c_scalar_fixture`, `scalar_skip_fixture`); their intended errors still fire.

All unit tests pass under `--mm:orc`. The regenerated header compiles cleanly under `gcc -std=c11 -Wall -Wextra` and `g++ -std=c++17`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,23 @@ All notable changes to this project are documented in this file.
 ## [Unreleased]
 
 ### Added
+- The `abi = c` C header now declares the event-listener ABI that
+  `declareLibrary` always exports (`<lib>_add_event_listener` /
+  `<lib>_remove_event_listener`) and the `FFICallBack` typedef they take, so a
+  consumer needs no hand-written header. The typed listener machinery for
+  `{.ffiEvent.}` is still unsupported under `abi = c`.
+- The `abi = c` C header is self-contained: it emits the `<stdint.h>` /
+  `<stddef.h>` includes, the `NIMFFI_RET_*` status codes, and short
+  `#ifndef`-guarded `RET_*` aliases for consumers that use the unprefixed names.
+- Each `{.ffi.}` proc's `##` doc comment reaches the generated C header as a
+  `/** ... */` block above the declaration and its wrapper.
+- `declareLibrary` accepts the `ABIFormat` enum for `defaultABIFormat`, so
+  `defaultABIFormat = ABIFormat.C` compiles alongside the `"c"` string.
+- `declareLibrary` takes an optional `headerBanner` argument, stamped as a
+  `//`-comment block at the top of every generated C header.
+- `genBindings()` fails compilation when a library declares an `{.ffiCtor.}` but
+  no `{.ffiDtor.}`, so the context a constructor builds always has a way to be
+  released.
 - `{.ffiEvent.}` now accepts multiple parameters. The macro synthesises and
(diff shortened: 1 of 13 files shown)


fcecin (approved):
LGTM

Issue #145, opened 2026-08-03
Security & memory-safety audit of master @ aad9374 — 11 findings (3 High) + test-first remediation plan


NagyZoltanPeter:
Security and memory-safety audit of the runtime FFI surface at `master` @ aad9374 ("Integrate the 0.2 line into master: `{.ffi.}` shape router, `{.ffiExport.}`, context recycling", #142).

**11 findings — 3 High, 2 Medium, 6 Low.** The three High findings share a root cause: the context-recycling feature returns a pool slot to service without fully draining it, fully resetting it, or re-identifying it.

Toolchain: Nim 2.2.4. Method: manual read of the full runtime surface plus the generated-code macros, in three independent passes; every finding re-verified against source. Two claims from earlier passes were refuted or downgraded and are recorded in the *Corrections* section rather than dropped.

---

## Scope

**In scope (runtime attack surface):** `ffi/alloc.nim`, `ffi/cbor_serial.nim`, `ffi/ffi_context.nim`, `ffi/ffi_context_pool.nim`, `ffi/ffi_events.nim`, `ffi/ffi_handles.nim`, `ffi/ffi_request_queue.nim`, `ffi/ffi_thread.nim`, `ffi/ffi_thread_request.nim`, `ffi/ffi_types.nim`, `ffi/event_thread.nim`, `ffi/internal/*.nim`.

**Out of scope:** `ffi/codegen/*.nim` — compile-time binding generators that only `writeFile` into an output directory. No runtime attack surface. Confirmed no `execCmd`/`startProcess`/`osproc`/network use anywhere under `ffi/` (only `os.getEnv("NO_COLOR")` at `logging.nim:93`).

**Threat model.** The C caller is semi-trusted (it is the embedding host), but the library should not corrupt memory or crash on malformed input, and must not leak state across context lifecycles. Hosts commonly forward network-derived data into FFI payloads, so payload bytes are treated as attacker-influenced.

## Summary

| ID | Finding | Severity | mm impact |
|----|---------|----------|-----------|
| SEC-01 | `recycleContext` discards its drain result and tears down unconditionally | **High** | both |
| SEC-02 | Handle registry never cleared on recycle → leak + stale-handle resolution | **High** | both (UAF worse under refc) |
| SEC-03 | Pooled context pointer has no generation tag → ABA across recycle | **High** | both |
| SEC-04 | `abi = c` CWire unpack trusts caller `_len` / `cstring` unconditionally | Medium | both |
[...]

PR #149, opened 2026-08-11
fix(ffi): dispatch events off a snapshot so a listener can mutate the registry

gmelodie:
## Summary

Follow-up on the findings of #145 (SEC-05). A listener callback that called `<lib>_remove_event_listener` deadlocked the event thread.

`dispatchToListeners` held `reg.lock` across the callbacks, and `Lock` is a plain pthread mutex, so a listener that dropped itself waited on the thread that was running it. The dispatch now snapshots the listeners, releases the lock and then calls them.

The lock did carry one guarantee worth keeping: after a remove returned, no callback of that listener could still run, so the host could free its `userData`. A bare snapshot trades the deadlock for a use-after-free, so the registry now counts the deliveries in flight and `removeEventListener` waits them out. A thread inside a dispatch skips the wait, which is the case that used to hang; the depth lives in a `{.threadvar.}`, so the count stays correct even if a second thread ever dispatches. `clearListeners` and `removeAllEventListeners` take the same wait, so a recycle cannot return while a callback runs.

This also resolves the contradiction the audit recorded: `removeEventListener` claimed "safe from inside a dispatch; the in-flight snapshot still delivers once" while the dispatch used no snapshot. The docstring is now true, and three further comments that named `reg.lock` as the hazard are corrected.

## Affected Areas

`ffi/ffi_events.nim` gains `beginDispatch` / `endDispatch` / `awaitDispatch` on `FFIEventRegistry`, plus a `Cond` and a delivery count. `ffi/event_thread.nim` uses them in `dispatchToListeners`. The C, C++ and Rust generators emit a doc comment on the remove entry point and on the wrapper that releases the listener box. The rest is comments and the new test job.

## Impact on Library Users

The C ABI does not change. The generated bindings carry the new doc comment, so the checked-in ones are regenerated; that diff is comments only.

A listener may now call `<lib>_remove_event_listener` from inside its own callback, and gets one last delivery from the snapshot in flight. A remove from any other thread blocks until that delivery ends, so the `userData` of a listener is safe to free as soon as the remove returns.

A remove from inside a callback cannot wait for its own dispatch, so it returns at once. If it drops a different listener, the snapshot in flight can still deliver to that listener, and the `userData` of that listener must stay alive until the dispatch ends. The generated wrappers (`<lib>_ctx_remove_event_listener` in C, `removeEventListener` in C++, `remove_event_listener` in Rust) release the box that holds the handler, so call them from inside a callback only for the listener that runs. The header documentation and the changelog carry the same warning.

Under `--mm:refc` a listener may only remove, not add. A registration grows the registry, and the event thread must not allocate into the thread-local heap that owns it.

## Risk Assessment

The wait keeps the old cross-thread behaviour, so a wedged listener blocks a remove exactly as it did before. It cannot deadlock a dispatching thread, because such a thread never waits.

`FFIEventRegistry` gains a `Cond` beside its `Lock`; both are created in `initEventRegistry` and destroyed in `deinitEventRegistry`.

## Tests

New `tests/unit/test_event_listener_reentrancy.nim`. A listener drops itself from inside its callback, and a second test blocks a callback while another thread removes the listener, to prove the remove waits.

The file arms a watchdog that quits the process, because the unpatched behaviour is a hang and not a failed check. On `b6c17dc` both tests hit the watchdog and exit 1, under orc and under refc alike. With the fix the full unit suite passes under both memory models, and under `NIM_FFI_SAN=tsan`.

A `Listener Reentrancy` job joins the cross-OS unit matrix in `ci.yml`. Without it the file would run only in the two Linux sanitizer jobs, and this fix rests on pthread mutex and condvar behaviour.

## References

Issue #145 (SEC-05). Follows #148, which covers SEC-01, SEC-02, SEC-03, SEC-06, SEC-09, SEC-10 and SEC-11.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,17 @@ jobs:
       nim-versions:   ${{ needs.versions.outputs.nim-versions }}
       nimble-version: ${{ needs.versions.outputs.nimble }}
 
+  listener-reentrancy:
+    # The dispatch barrier is a pthread condvar; cover the OSes whose mutex and
+    # condvar behaviour differ, not only the Linux sanitizer jobs.
+    name: Listener Reentrancy
+    needs: versions
+    uses: ./.github/workflows/test.yml
+    with:
+      test: test_event_listener_reentrancy
+      nim-versions:   ${{ needs.versions.outputs.nim-versions }}
+      nimble-version: ${{ needs.versions.outputs.nimble }}
+
   c-wire:
     # The `c`-ABI cwire codec is layout-/allocator-sensitive (malloc/free, flat
     # struct packing, allocShared for seq/Option), so cover it across the full
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -152,6 +152,16 @@ All notable changes to this project are documented in this file.
   through the callback. The check is emitted only for `ref` library types, and
   it runs on the FFI thread behind any queued constructor, so a host that issues
   a call without awaiting the create callback is unaffected.
+- A listener callback that calls `<lib>_remove_event_listener` or
+  `<lib>_add_event_listener` no longer deadlocks the event thread. The dispatch
+  held the non-reentrant registry lock across the callbacks, so a listener that
+  dropped itself waited on the thread it ran on. The dispatch now calls the
+  listeners off a snapshot with the lock released, and the removed listener gets
+  one last delivery from that snapshot. A remove from another thread still waits
+  the delivery out, so the `userData` of a listener stays safe to free as soon as
+  the remove returns. Under `--mm:refc` a listener may only remove, not add: a
+  registration grows the registry, and the event thread must not allocate into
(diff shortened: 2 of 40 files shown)


fcecin (approved):
LGTM. Fable 5 review flagged a documentation (API contract) concern, FWIW.

```
  **Observation.** A listener callback can remove a different listener during a dispatch. The remove returns immediately, because a thread cannot wait for its own dispatch. The removed listener
  can still receive one delivery from the snapshot of that dispatch.

  **Risk.** The changelog says userData is safe to free as soon as the remove returns. That is true only for a remove from another thread. If a listener removes a different listener and the host
  then frees that userData, the last delivery writes to freed memory.

  **Recommendation.** Add one warning to the changelog and to the header documentation: "Do not free the userData of a listener that you remove from inside a listener callback. The current
  dispatch can deliver to it one more time." Optional follow-up, not a blocker: return a distinct code from remove for this case, so the host can detect it.
```

PR #150, opened 2026-08-11
fix(ffi): bound the request ingress queue and the request payload

gmelodie:
## Summary

Follow-up work on the findings of #145. Request ingress had no limit of any kind: neither on how many requests one producer could leave queued, nor on how large a single payload could be.

- **The ingress queue took every request offered.** A producer faster than the FFI thread — or any producer at all once that thread is wedged — grew memory without bound, while the event queue next to it has been bounded at 1024 all along. Each queue now holds at most `RequestQueueDepth` requests, counted under the lock it already takes, and `pushRequest` answers `QueueFull` instead of appending. Ingress is sharded over 16 queues, so a context holds at most 16384 requests.
- **No request carried a maximum size.** `sendRequestToFFIThread` now rejects a payload over `MaxRequestPayloadBytes` before it reaches the queue or the decoder.

Both limits are compile-time defines: `-d:ffiRequestQueueDepth` (1024) and `-d:ffiMaxRequestPayloadBytes` (8 MiB).

## Impact on Library Users

A submit can now fail for two new reasons. Both come back as the error the submit already returns, which the generated entry points report as `RET_ERR` through the callback, so a host that stays under the limits sees no change and the C ABI is untouched.

The payload cap measures the buffer the request owns. On the `abi = c` path that buffer is the packed wire struct, so the memory its fields point at stays uncounted — that path trusts its caller, as it already does for every length it is handed.

`pushRequest` returns `PushOutcome` rather than a `bool`, and `RequestQueueDepth` is re-exported from `ffi_context`. Both are internal to the runtime; no generated binding changes.

## Tests

New `tests/unit/test_ffi_ingress_limits.nim`, with a sibling `.cfg` cutting the caps to 4 requests and 1024 bytes so the cases cost nothing. A handler blocks the FFI thread outright, so the queue fills deterministically: `RequestQueueDepth` submits succeed, the next one errors, and once the handler is released and the queue drains a submit is accepted again — that last check is what catches a depth counter that never resets. A payload one byte over the cap errors while a small one still passes.

Both cases were run with the two new guards stubbed out and both fail there. The suite passes under `--mm:orc` and `--mm:refc`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,21 @@ All notable changes to this project are documented in this file.
   returns the library type builds today and exports a working C symbol, so a
   router would silently give it the ctor ABI instead.
 
+### Changed
+- **A submit now has two limits, and fails instead of accepting without bound.**
+  The ingress queue took every request a producer offered, so a producer faster
+  than the FFI thread — or any producer once that thread is wedged — grew memory
+  without bound, and no request carried a maximum size. A submit is rejected once
+  its ingress queue holds `-d:ffiRequestQueueDepth` (1024) requests, or once its
+  payload passes `-d:ffiMaxRequestPayloadBytes` (8 MiB). Ingress is sharded over
+  16 queues, so a context holds at most 16384 requests. The payload cap measures
+  the buffer the request owns: on the `abi = c` path that is the packed wire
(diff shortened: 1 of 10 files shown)


fcecin (approved):
LGTM

PR #151, opened 2026-08-17
fix(ffi): quarantine a context whose {.ffiDtor.} did not finish

NagyZoltanPeter:
## Summary

A pooled context is recycled by handing its slot — **including the live FFI and event thread pair and their chronos dispatcher** — to the next owner. `recycleContext` decided that was safe by looking at the request drain alone: `runTeardown` returned `void` and swallowed both its `withTimeout` expiry and any raise. So a `{.ffiDtor.}` cut short by `TeardownTimeout` (10 s), or one that raised, still had its library freed and its slot released, and `requestRecycle` reported **success**.

The gap that leaves: whatever async work the abandoned teardown never cancelled keeps running on that thread. It is not a theoretical window — the new test reproduces the whole chain on `master`.

### The failure chain, measured

A library spawns a loop (`asyncSpawn`), its dtor overruns the timeout and never cancels it. On the unpatched runtime:

| What the host sees | Reality |
| --- | --- |
| `<lib>_ctx_destroy` → `RET_OK` | the dtor body was cancelled mid-way; only a log line says so |
| the context is gone | `ctx[].myLib` is **nil** — the library was freed *under* the still-running loop. Reading it segfaults |
| a fresh context | the **same slot**, served by the **same thread id** as the orphan (`41284909` in the run) |
| its own event stream | its listener received **3 events belonging to the previous owner** |
| its own library | `myLib.canary` came back as **`233492702` = `0x0DEAD0DE`** — the orphan's sentinel, written through its stale `ptr T`, inside the *next owner's* library object |
| — | under `asan+ubsan` the same write hits the allocator's free list and **aborts the process**: `member access within misaligned address 0x00000dead0de for type 'FreeCell'` |

The last line is worse than a use-after-free: Nim's shared allocator hands out sub-blocks of large `malloc` chunks, so ASan cannot see the per-object free — the corruption surfaces when the allocator follows the sentinel as a free-list pointer.

### Why the runtime cannot clean up after a cut-short teardown

There is no "cancel everything on this thread" to reach for:

| Fact | Where |
| --- | --- |
| the pending-future list is per-thread, so the scope would be right | `chronos/futures.nim:118` — `var futureList* {.threadvar.}` |
| enumeration exists (`pendingFutures`, `pendingFuturesCount`)… | `chronos/internal/asyncengine.nim:1261` |
| …but only under `-d:chronosFutureTracking` (default off) | `chronos/config.nim:47` |
| `cancelSoon(FutureBase)` is public, so cancelling an enumerated future would work | `chronos/internal/asyncfutures.nim:1000` |
| the list carries **no ownership tag** — the FFI loop's own futures are in it | cancelling untargeted would kill the runtime |
| a cancel is cooperative regardless | a future that swallows `CancelledError` never dies |

Reuse is only safe when the previous owner is provably gone from the thread, and the only proof available is its teardown having run to the end.

## What changed

A recycle releases the slot **only** when the drain succeeded *and* the teardown either completed or had nothing to do (`Skipped`: no hook, or no library because the ctor never landed). Anything else **quarantines**:

- the slot stays claimed for the life of the process, so `createFFIContext` never offers it again;
- the library is **kept alive** — the orphans still hold pointers into it, and freeing it is what turned a leak into corruption;
- the terminal `RecycleFailed` carries a `RecycleFailure` reason (`DrainTimeout`, `TeardownTimeout`, `TeardownRaised`, `CallerAbandoned`), which `requestRecycle` returns as its error. An enum and not a message: a `string` field on a pooled context is written by the FFI thread and read by the host thread, which crosses thread-local heaps under refc.

Also:

- **The late-completion hole is closed.** A teardown that ignores cancellation outlasts the caller's own `RecycleWaitTimeout`; that caller was already told the recycle failed, so `requestRecycle` marks the context abandoned and the recycle quarantines instead of silently handing the slot back later.
- **Joining the threads is the one way out.** `destroyFFIContext` still stops and joins the pair — the orphans die with the dispatcher — and `releaseSlot` now clears the quarantine so the slot rebuilds cleanly.
- **`eventQueueStuck` is cleared on the recycle reset.** It is set on event-queue overflow and cleared only in `initContextResources`, which a reused slot skips, so one overflow left the slot rejecting every request of every later owner — the constructor included, which made `createFFIContext` hand out the same poisoned slot again and again, bricking context creation for the process.
- **Every path that stops serving without a completed teardown logs at `error`**: drain timeout, teardown timeout, teardown raised, caller's wait expired, thread-exit timeout in `destroyFFIContext` and in `destroyStaticFFIContext`, and pool exhaustion once any slot has been quarantined. A teardown skipped because there was nothing to tear down logs at `debug` — that is not a failure.
- `FFIContextPool.quarantinedSlots()` reports the lost capacity, and the pool-exhausted error names the count.

## Impact on Library Users

**Behavioural break.** `<lib>_ctx_destroy` now returns `RET_ERR` where it returned `RET_OK`: when the `{.ffiDtor.}` overran its timeout or raised. On that error the context and its library are leaked deliberately and their callbacks can still fire, so the host must keep the `userData` of in-flight requests alive. The C ABI itself is unchanged and no generated binding changes (`nimble check_bindings` clean).

**The dtor contract, now written down** (README → *The teardown contract*): a `{.ffiDtor.}` must cancel **and await** everything it spawned. A quarantine costs one of the 32 slots for the life of the process, so a library that habitually overruns its teardown will exhaust the pool — loudly, with the count named, rather than by silently corrupting the next owner.

**Nim API:** `RecycleFailure` + `reason()`, `FFIContextPool.quarantinedSlots()`, and `FFIContext` gains `recycleFailure` / `recycleAbandoned`.

## Risk Assessment

The change only ever *withholds* a slot, so the failure direction is capacity, not safety. The hard stop is a clean, attributable `pool exhausted (max 32 contexts, N quarantined by a failed teardown)` instead of a diffuse corruption far from the cause.

Note that both the old and the new behaviour leak the orphan **threads** — quarantine does not make them stop. What it prevents is a *second* owner sharing them.

The `recycleAbandoned` flag races the recycle's own completion by design; losing that race can only release a slot whose teardown did complete, so the unsafe direction is unreachable.

One pre-existing edge this makes more visible: `destroyFFIContext` on a quarantined context re-enters the `{.ffiDtor.}`, because the shutdown path ends in `runTeardown` and `libReady`/`myLib` are still set (the reset is skipped on quarantine). Worth a follow-up guard; it is not introduced here.

## Tests

New `tests/unit/test_ffi_dtor_orphan_reuse.nim`, with a sibling `.cfg` cutting `TeardownTimeout` to 300 ms and the drain rounds to 200 ms. An `asyncSpawn`ed loop that the dtor never cancels ticks, fires an event and writes a sentinel through its stale `ptr T`; per-incarnation counters keep an orphan leaked by one case from moving the numbers another case asserts on. Five cases:

1. a dtor that stops and awaits its offspring recycles cleanly and the slot **is** reused (the control — it runs first, while no orphan is alive);
2. a dtor cut short by the timeout fails the recycle, and the library is kept while the orphan keeps ticking;
3. the next owner gets **another** slot, on another thread, never sees the orphan's events, and its library keeps its own canary;
4. the abandoned library is not freed under the orphan — armed before the recycle with nothing allocating afterwards, this is the case the ASan/UBSan job reports as allocator corruption on `master`;
5. a teardown that finishes *past* `RecycleWaitTimeout` still does not release the slot, and three later creates never land on it.

All five fail on the unpatched recycle path (case 1 passes, as it must); with the fix they pass under `--mm:orc`, `--mm:refc`, `NIM_FFI_SAN=asan-ubsan` and `NIM_FFI_SAN=tsan`. Full suite: `nimble test` green under both memory models.

`test_ffi_teardown`'s *"a teardown past TeardownTimeout still releases the slot"* asserted the old behaviour and is rewritten as *"…quarantines the slot"*.

## CI

Two gaps, both fixed here.

**The new file would have run almost nowhere.** `ci.yml` names one job per test file for the cross-OS matrix (3 OS × Nim versions × orc/refc), and that list covered 9 of the 43 unit test files. Everything else ran only inside the two Linux sanitizer jobs, which call `nimble test_sanitized` over every discovered file — so a new file gets no macOS, no Windows and no unsanitized run. The recycle/teardown family all turn on thread reuse, thread exit and dispatcher behaviour, which differ per OS, so each now has its own matrix job: the new orphan-reuse file plus `test_ffi_teardown`, `test_ffi_recycle_reset`, `test_ffi_claim_generation`, `test_ffi_ingress_limits`, `test_ctx_after_failed_ctor` and `test_event_thread`. Per-file matrix coverage goes 9 → 16.

**A new file could go unrun again.** `unit-tests-all` runs `nimble test` — all 43 files, orc and refc — as a catch-all.

Consequences, deliberately chosen:

- `unit-tests-all` is **Linux only**. The cost here is compiling ~40 files twice (measured 4:25 wall locally; expect 10–18 min on a runner), and the jobs above already carry the OS-sensitive files across the matrix.
- It is an **addition, not a replacement**. One aggregate job on the full matrix would serialise inside each cell and cost *more* wall clock than the current one-job-per-file design, which runs them in parallel.
- Nothing needed in `ffi.nimble`: `test.yml` compiles `tests/unit/<test>.nim` directly, and `nimble-job.yml` already runs an arbitrary nimble command on one OS. No changes to either reusable workflow.
- Every referenced file was verified to exist, and each new block is a clone of the existing `listener-reentrancy` / `check-bindings` shape.

## References

Issue #145 (SEC-01 was the drain half of this; this PR is the teardown half). Follows #147, which put the teardown on the recycle path, and #148, which reset the slot and made the host's context handle opaque.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,6 +112,72 @@ jobs:
       nim-versions:   ${{ needs.versions.outputs.nim-versions }}
       nimble-version: ${{ needs.versions.outputs.nimble }}
 
+  # The recycle/teardown family below all turn on thread reuse, thread exit and
+  # chronos dispatcher behaviour, which differ per OS. The Linux sanitizer jobs
+  # run every unit test, but they cannot speak for macOS or Windows.
+  dtor-orphan-reuse:
+    name: Dtor Orphan Reuse
+    needs: versions
+    uses: ./.github/workflows/test.yml
+    with:
+      test: test_ffi_dtor_orphan_reuse
+      nim-versions:   ${{ needs.versions.outputs.nim-versions }}
+      nimble-version: ${{ needs.versions.outputs.nimble }}
+
+  ffi-teardown:
+    name: Async Teardown
+    needs: versions
+    uses: ./.github/workflows/test.yml
+    with:
+      test: test_ffi_teardown
+      nim-versions:   ${{ needs.versions.outputs.nim-versions }}
+      nimble-version: ${{ needs.versions.outputs.nimble }}
+
+  recycle-reset:
+    name: Recycle Reset
+    needs: versions
+    uses: ./.github/workflows/test.yml
+    with:
+      test: test_ffi_recycle_reset
+      nim-versions:   ${{ needs.versions.outputs.nim-versions }}
+      nimble-version: ${{ needs.versions.outputs.nimble }}
+
+  claim-generation:
+    name: Claim Generation
+    needs: versions
+    uses: ./.github/workflows/test.yml
+    with:
+      test: test_ffi_claim_generation
+      nim-versions:   ${{ needs.versions.outputs.nim-versions }}
+      nimble-version: ${{ needs.versions.outputs.nimble }}
+
+  ingress-limits:
+    name: Ingress Limits
+    needs: versions
+    uses: ./.github/workflows/test.yml
+    with:
+      test: test_ffi_ingress_limits
+      nim-versions:   ${{ needs.versions.outputs.nim-versions }}
+      nimble-version: ${{ needs.versions.outputs.nimble }}
+
+  ctx-after-failed-ctor:
+    name: Ctx After Failed Ctor
+    needs: versions
+    uses: ./.github/workflows/test.yml
+    with:
+      test: test_ctx_after_failed_ctor
+      nim-versions:   ${{ needs.versions.outputs.nim-versions }}
+      nimble-version: ${{ needs.versions.outputs.nimble }}
+
+  event-thread:
+    name: Event Thread
+    needs: versions
+    uses: ./.github/workflows/test.yml
+    with:
+      test: test_event_thread
+      nim-versions:   ${{ needs.versions.outputs.nim-versions }}
+      nimble-version: ${{ needs.versions.outputs.nimble }}
+
   cpp-e2e:
     # Codegen output doesn't vary with mm, so we matrix over OS and Nim only.
     # Windows runs MSVC by default and may surface codegen tweaks needed in
@@ -204,6 +270,19 @@ jobs:
     uses: ./.github/workflows/nimble-job.yml
     with:
       run: nimble check_bindings -y
+      nim-versions:   ${{ needs.versions.outputs.nim-versions }}
+      nimble-version: ${{ needs.versions.outputs.nimble }}
+
+  unit-tests-all:
+    # Catch-all so a new tests/unit file cannot go unrun: `nimble test` discovers
+    # every test_*.nim and runs it under both orc and refc. Linux only, because
+    # the compile of ~40 files twice is the cost here and the jobs above already
+    # carry the OS-sensitive ones across the matrix.
+    name: All unit tests (orc + refc)
+    needs: versions
+    uses: ./.github/workflows/nimble-job.yml
+    with:
+      run: nimble test -y
       nim-versions:   ${{ needs.versions.outputs.nim-versions }}
       nimble-version: ${{ needs.versions.outputs.nimble }}
 
--- a/ffi/ffi_context_pool.nim
+++ b/ffi/ffi_context_pool.nim
@@ -88,8 +88,8 @@ proc createFFIContext*[T](
       "does not finish costs a slot for the life of the process",
       quarantined = quarantined, max = MaxFFIContexts
   err(
-    "FFI context pool exhausted (max " & $MaxFFIContexts & " contexts, " &
-      $quarantined & " quarantined by a failed teardown)"
+    "FFI context pool exhausted (max " & $MaxFFIContexts & " contexts, " & $quarantined &
+      " quarantined by a failed teardown)"
   )
 
 proc isStaticCtx[T](pool: var FFIContextPool[T], ctx: ptr FFIContext[T]): bool =
--- a/ffi/ffi_context_pool.nim
+++ b/ffi/ffi_context_pool.nim
@@ -1,5 +1,5 @@
 import std/[atomics, sysatomics]
-import results
+import chronicles, results
 import ./ffi_context
 
 const MaxFFIContexts* = 32
@@ -45,7 +45,21 @@ proc releaseSlot[T](pool: var FFIContextPool[T], ctx: ptr FFIContext[T]) =
     if pool.contexts[i].addr == ctx:
       pool.initialized[i].store(false)
       break
+  # Joining the threads is the one way out of a quarantine: whatever a cut-short
+  # teardown left behind died with the thread, and the next claim rebuilds the
+  # pair. `initContextResources` would store this anyway; doing it here keeps
+  # `quarantinedSlots` from counting a slot that is no longer out of service.
+  ctx.lifecycle.store(CtxLifecycle.Active)
+  ctx.recycleFailure.store(RecycleFailure.None)
   ctx.releaseClaim()
+
+proc quarantinedSlots*[T](pool: var FFIContextPool[T]): int =
+  ## Slots taken out of service by a recycle that could not prove the previous
+  ## owner was done with the thread. Only the recycle handler writes
+  ## `RecycleFailed`, and only a full destroy clears it, so counting it is exact.
+  for i in 0 ..< MaxFFIContexts:
+    if pool.contexts[i].lifecycle.load() == CtxLifecycle.RecycleFailed:
+      inc result
 
 proc createFFIContext*[T](
     pool: var FFIContextPool[T]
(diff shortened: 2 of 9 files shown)


ffi/ffi_context_pool.nim
```diff
@@ -45,8 +45,22 @@ proc releaseSlot[T](pool: var FFIContextPool[T], ctx: ptr FFIContext[T]) =
...
+  ctx.recycleFailure.store(RecycleFailure.None)
   ctx.releaseClaim()
 
+proc quarantinedSlots*[T](pool: var FFIContextPool[T]): int =
+  ## Slots taken out of service by a recycle that could not prove the previous
+  ## owner was done with the thread. Only the recycle handler writes
+  ## `RecycleFailed`, and only a full destroy clears it, so counting it is exact.
+  for i in 0 ..< MaxFFIContexts:
+    if pool.contexts[i].lifecycle.load() == CtxLifecycle.RecycleFailed:
+      inc result
```
gmelodie: Avoid using `result` as per logos coding guidelines

tests/unit/test_ffi_teardown.nim
```diff
@@ -46,6 +46,15 @@ proc encodedPtr(bytes: var seq[byte]): ptr byte =
   else:
     cast[ptr byte](addr bytes[0])
 
+proc waitSlotFree[T](ctx: ptr FFIContext[T]) =
```
gmelodie: This proc is pasted three times
- `tests/unit/test_ffi_teardown.nim`
- `tests/unit/test_ffi_dtor_orphan_reuse.nim`
- `tests/unit/test_ffi_recycle_reset.nim`

ffi/ffi_thread.nim
```diff
@@ -198,34 +211,64 @@ proc resetForNextOwner[T](ctx: ptr FFIContext[T], ongoing: ptr seq[Future[void]]
...
+    # must not come back. Racing that flag is benign: losing it can only release
+    # a slot whose teardown did complete.
+    if failure == RecycleFailure.None and ctx.recycleAbandoned.load():
+      failure = RecycleFailure.CallerAbandoned
+    if failure != RecycleFailure.None:
+      ctx.recycleFailure.store(failure)
       ctx.lifecycle.store(CtxLifecycle.RecycleFailed)
+      error "context quarantined; the pool slot and its threads are leaked, the " &
+        "library is kept alive and its callbacks can still fire",
+        reason = failure.reason(), cause = $failure
```
gmelodie: I think this could be a separate `proc`

.github/workflows/ci.yml
```diff
@@ -112,6 +112,70 @@ jobs:
...
+
+  event-thread:
+    name: Event Thread
+    needs: versions
+    uses: ./.github/workflows/test.yml
+    with:
+      test: test_event_thread
+      nim-versions:   ${{ needs.versions.outputs.nim-versions }}
+      nimble-version: ${{ needs.versions.outputs.nimble }}
+
```
gmelodie: 1. Let's merge this into one. Claude has the numbers:

> GitHub caps org concurrency, so 84 jobs queue behind each other anyway, and the setup overhead is paid 84 times. A single job whose run step loops over the 7 files pays setup 12 times and adds ~6 extra compile+run cycles per cell. Total CPU-minutes drop about 6×, and under a loaded queue wall clock drops too.

2. I think we don't need to test multiple nim/nimble versions, just the latest stable. This is an OS test, so nim variability shouldn't matter much.

3. We can also drop `ubuntu` from the tested OSes, since this already runs in the linux catch all below.

.github/workflows/ci.yml
gmelodie, line 8: let's add a concurrency clause here too, like we do in `linters.yml`
```suggestion
concurrency:
    group: ${{ github.workflow }}-${{
  github.event.pull_request.number || github.ref }}
    cancel-in-progress: ${{ github.event_name == 'pull_request'
  }}
  
```

tests/unit/test_ffi_dtor_orphan_reuse.nim
```diff
@@ -0,0 +1,348 @@
+## `TeardownTimeout` cuts a `{.ffiDtor.}` short and abandons the async work it did not cancel; the orphans still run on the FFI thread's dispatcher, chronos cannot cancel every future of a thread, so the slot must quarantine instead of serve the next owner.
```
gmelodie: This file has a lot of duplicated harnessing with `test_ffi_teardown.nim`. Almost half the lines here are already present in `test_ffi_teardown`. `encodedPtr` also appears everywhere. 

1. merge this file with `test_ffi_teardown.nim`
2. add a `tests/unit/helpers.nim` for the shared helpers.

these two will help reduce the footprint of this pr by a lot

gmelodie (commented):
> This PR suggest that when a tear-down flow is not successful somehow - like time outs - instead of recycling the context in the available pool, let's mark it as quarantined.
This can prevent re-using a thread possible running chronos async loops and dangling objects.

Good idea!

gmelodie (approved):

